PR #1, opened 2020-03-23
OpenGLDevice entry point loading

TheSpydog:
Submitting as a PR because there's a lot to look at here. The headlines:
* Added GL entry point loading logic (with fallbacks) for OpenGLDevice.
* Added `visualc` directory for Windows.
* Added part of the initialization logic for OpenGLDevice. (Essentially porting over the .cs constructor.)
* Added a simple test to ensure everything is loading appropriately.

Note that I didn't add the test to the Makefile because I suck at writing Makefiles.

Also worth noting that in `visualc`, FNA3D is being compiled as C++ to work around some MSVC C99 limitations... not sure if we should stick with that or make some changes to the code itself.


flibitijibibo:
The vcxproj for FNA3D appears to be a managed C++ project, we’ll want to rip out that part of the project file... I’ll run this through -Wpedantic -std=c90 tomorrow to deal with all the errors.

TheSpydog:
Okay, I think I ripped out all the managed stuff from the FNA3D vcxproj. Also added Release configs for both projects.

flibitijibibo:
Squashed everything in after doing a pass through everything. The support bools will probably be the biggest divergence from the existing FNA OpenGLDevice, but it's probably the only way to keep the GL support flags sane since we don't have exception handlers to be lazy with. I documented the extensions and their bools as best as possible, hopefully it'll be easy to squeeze them into the C translation.

PR #3, opened 2020-03-25
Quick and dirty CMakeLists

cybik:
VERY quick, functionnal-ish CMakeLists file with one pre-configured vscode settings file as a baseline.


flibitijibibo:
I don't think we'll use this unless we have a life-or-death reason; CMake is usually for configurable builds and I have no intention of adding configurations to FNA3D (other than debug/release maybe). Use it locally if you want though!

cybik:
It's not life-or-death, but VScode doesn't like Makefiles.

PR #6, opened 2020-03-26
OpenGL CreateTexture implementations

cybik:


flibitijibibo:
Will need to do a VS2010 pass for the variable declarations, but otherwise this is good!

PR #12, opened 2020-03-26
OpenGL ResolveTarget and SetRenderTargets

TheSpydog:
I took the approach of just stuffing all the possibly relevant information into a single struct. That way we don't have to deal with managed Texture references. I haven't written out the FNA code for this yet, but I think it should be fairly simple.

### DIFF
--- a/include/FNA3D.h
+++ b/include/FNA3D.h
@@ -411,14 +411,13 @@ typedef struct FNA3D_RenderTargetBinding
 	/* Texture */
 	FNA3D_SurfaceFormat format;
 	int32_t levelCount;
-	uint32_t textureHandle;
+	FNA3D_Texture *texture;
 
 	/* IRenderTarget */
 	int32_t width;
 	int32_t height;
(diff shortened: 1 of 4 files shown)


include/FNA3D.h
```diff
@@ -401,6 +401,34 @@ typedef struct FNA3D_PresentationParameters
...
+	int32_t width;
+	int32_t height;
+	FNA3D_RenderTargetUsage renderTargetUsage;
+	uint32_t depthStencilBufferHandle;
+	uint32_t colorBufferHandle;
+	FNA3D_DepthFormat depthStencilFormat;
+	int32_t multiSampleCount;
+
+	/* RenderTarget2D/Cube */
+	uint8_t isContentLost;
```
TheSpydog: Do we need this...?

flibitijibibo: Probably not, at least not until we have evidence that any device would ever lose a target in 2020... To that end, I would cut any variables in here that are not in active use, until we know for sure that they're needed.

This looks good to me, how does it look on the managed side?

PR #13, opened 2020-03-27
WIP: VulkanRenderer

thatcosmonaut:
Vulkan tasks are being tracked here: https://github.com/thatcosmonaut/FNA3D/issues/9

Contributions welcome! There's some good easy-level tasks available for graphics newbies who want to contribute.


src/FNA3D_Driver_Vulkan.c
```diff
@@ -187,9 +188,13 @@ struct RenderPassHashMap
 typedef struct FramebufferHash
 {
 	VkImageView colorAttachmentViewOne;
+	VkImageView colorMultiSampleAttachmentViewOne;
```
TheSpydog, line 191: I'm pretty sure you could make colorAttachments and colorMultiSampleAttachments into fixed-length arrays, since it should get packed the same anyway. Might be a bit easier than wrangling 8 different struct members. :)

thatcosmonaut: I was playing it safe because I didn't know if the data would get packed correctly in an array, thanks for the heads up!

flibitijibibo:
Squashed and merged!

https://github.com/FNA-XNA/FNA3D/commit/50ce672c082374e8014b2477796c45dbda1491cd

PR #15, opened 2020-03-29
[WIP] Implement threaded AddDispose calls

TheSpydog:


flibitijibibo:
Going to merge this now - supposedly there is a memory issue in here but this is at least a valid copypaste of what we have in the C# version.

PR #19, opened 2020-04-10
WIP: D3D11Renderer

TheSpydog:
This is a D3D11 driver for FNA3D. While D3D is _technically_ already supported via the OpenGL backend and ANGLE, having a native backend should improve performance and remove the ANGLE dependency on Windows and Xbox.

All that's left is fixing Katelyn's rendering tests, then this is good to go!


flibitijibibo:
Did a review of the whole file as of the latest commit. Everything not already marked with a TODO or FIXME looks good. Once all this is done we'll review the MojoShader stuff next (since that's a whole other oddity), after which we'll finally get this in master!

flibitijibibo:
Pushed the last of the renderer changes that I had in mind. The process from here:

- Review the MojoShader changes (read: FEZ crash fix)
- Merge MojoShader first, update FNA3D master's submodule
- Squash this PR down to one commit that doesn't modify the submodule
- Merge!

PR #29, opened 2020-07-03
[WIP] Vulkan Threading Support

flibitijibibo:


flibitijibibo:
This is pretty out of date and doesn't work at all, so we'll just drop this... ongoing discussion is happening in the Discord server.

PR #32, opened 2020-07-23
[VULKAN] Optimize Pipeline Lookups

thatcosmonaut:
Reduces the number of times we have to look up a pipeline by only re-fetching pipelines if some pipeline-related state has changed.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -6519,28 +6519,15 @@ static void VULKAN_SetBlendState(
 ) {
 	VulkanRenderer *renderer = (VulkanRenderer*) driverData;
 
-	if (renderer->blendState.blendFactor.r != blendState->blendFactor.r ||
-		renderer->blendState.blendFactor.g != blendState->blendFactor.g ||
-		renderer->blendState.blendFactor.b != blendState->blendFactor.b ||
-		renderer->blendState.blendFactor.a != blendState->blendFactor.a)
-	{
-		VULKAN_SetBlendFactor(
-			driverData,
-			&blendState->blendFactor
-		);
+	VULKAN_SetBlendFactor(
+		driverData,
+		&blendState->blendFactor
+	);
 
-		renderer->needNewPipeline = 1;
-	}
-
-	if (renderer->blendState.multiSampleMask != blendState->multiSampleMask)
-	{
-		VULKAN_SetMultiSampleMask(
-			driverData,
-			blendState->multiSampleMask
-		);
-
-		renderer->needNewPipeline = 1;
-	}
+	VULKAN_SetMultiSampleMask(
+		driverData,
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_Vulkan.c
```diff
@@ -3934,17 +3938,29 @@ static VkPipeline VULKAN_INTERNAL_FetchPipeline(VulkanRenderer *renderer)
 
 static void VULKAN_INTERNAL_BindPipeline(VulkanRenderer *renderer)
 {
-	VkPipeline pipeline = VULKAN_INTERNAL_FetchPipeline(renderer);
+	VkShaderModule vertShader, fragShader;
+	MOJOSHADER_vkGetShaderModules(&vertShader, &fragShader);
 
-	if (pipeline != renderer->currentPipeline)
+	if (renderer->needNewPipeline ||
```
flibitijibibo, line 3944: Tabs for ( and )

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6499,29 +6519,50 @@ static void VULKAN_SetBlendState(
...
 		driverData,
 		&blendState->blendFactor
 	);
+
 	VULKAN_SetMultiSampleMask(
 		driverData,
 		blendState->multiSampleMask
 	);
+
+	SDL_memcpy(&renderer->blendState, blendState, sizeof(FNA3D_BlendState));
```
flibitijibibo, line 6532: Do we need to do an SDL_memcmp and request a new pipeline for changes here?

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6499,29 +6519,50 @@ static void VULKAN_SetBlendState(
...
 	FNA3D_DepthStencilState *depthStencilState
 ) {
 	VulkanRenderer *renderer = (VulkanRenderer*) driverData;
-	SDL_memcpy(
-		&renderer->depthStencilState,
-		depthStencilState,
-		sizeof(FNA3D_DepthStencilState)
-	);
+
+	if (renderer->depthStencilState.stencilMask != depthStencilState->stencilMask ||
```
flibitijibibo, line 6541: Since we're storing with memcpy we can probably replace all this with memcmp to keep it short.

thatcosmonaut: Oh, i didn't know about memcmp. Neat! I can rewrite some of these to use that func instead

src/FNA3D_Driver_Vulkan.c
thatcosmonaut: the comparisons and pipeline checks happen in the above functions, not sure if memcmp is worth doing before the copy?

thatcosmonaut: Oh wait, I see, yes, this is an error actually

PR #40, opened 2020-08-24
When creating a D3D11 debug device fails try creating a regular one

kg:
For users without the necessary debug bits on their machine, creating a D3D11 device with the debug flag on fails with a specific cryptic HRESULT. So this PR adds a mechanism where if we fail to create a device in debug mode we try making a non-debug one before finally giving up.

### DIFF
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -4609,7 +4609,7 @@ try_create_device:
 	{
 		/* Creating a debug mode device will fail on some systems due to the necessary
 		 * debug infrastructure not being available. Remove the debug flag and retry. */
-		FNA3D_LogWarn("Creating device in debug mode failed with error %08X. Trying non-debug.");
+		FNA3D_LogWarn("Creating device in debug mode failed with error %08X. Trying non-debug.", res);
 		flags ^= D3D11_CREATE_DEVICE_DEBUG;
 		debugMode = 0;
 		goto try_create_device;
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_D3D11.c
```diff
@@ -4602,6 +4604,17 @@ static FNA3D_Device* D3D11_CreateDevice(
...
 		&renderer->context
 	);
+
+	if (FAILED(res) && debugMode)
+	{
+		/* Creating a debug mode device will fail on some systems due to the necessary
+		 * debug infrastructure not being available. Remove the debug flag and retry. */
+		FNA3D_LogWarn("Creating device in debug mode failed with error %08X. Trying non-debug.");
+		flags ^= D3D11_CREATE_DEVICE_DEBUG;
+        debugMode = 0;
```
flibitijibibo, line 4614:
```suggestion
		debugMode = 0;
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -4602,6 +4604,17 @@ static FNA3D_Device* D3D11_CreateDevice(
 		&renderer->featureLevel,
 		&renderer->context
 	);
+
+	if (FAILED(res) && debugMode)
+	{
+		/* Creating a debug mode device will fail on some systems due to the necessary
+		 * debug infrastructure not being available. Remove the debug flag and retry. */
+		FNA3D_LogWarn("Creating device in debug mode failed with error %08X. Trying non-debug.");
```
flibitijibibo, line 4612:
```suggestion
		FNA3D_LogWarn("Creating device in debug mode failed with error %08X. Trying non-debug.", res);
```

PR #43, opened 2020-08-25
Vulkan Sampler Descriptor Optimization

thatcosmonaut:
This patch reduces the number of descriptor set fetches by returning descriptor set references immediately at draw call time.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -2664,10 +2664,7 @@ static uint8_t VULKAN_INTERNAL_AllocateDescriptorSets(
 ) {
 	VkResult vulkanResult;
 	uint32_t i;
-	VkDescriptorSetAllocateInfo descriptorSetAllocateInfo =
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_Vulkan.c
```diff
@@ -2726,20 +2643,37 @@ static uint8_t VULKAN_INTERNAL_CreateDescriptorPoolAndSets(
...
+static uint8_t VULKAN_INTERNAL_AllocateDescriptorSets(
+	VulkanRenderer *renderer,
+	VkDescriptorPool descriptorPool,
+	VkDescriptorSetLayout descriptorSetLayout,
+	uint32_t descriptorSetCount,
+	VkDescriptorSet *descriptorSetArray
+) {
+	VkResult vulkanResult;
+	uint32_t i;
+	VkDescriptorSetAllocateInfo descriptorSetAllocateInfo =
```
flibitijibibo, line 2667: Initializer

PR #48, opened 2020-09-08
Fix issue with mixing up the hash and bindingsIndex in D3D11_INTERNAL…

bartwe:
…_FetchBindingsInputLayout

Fix issue with inputLayoutCache invalidation when effect gets deleted, pointer was getting recycled by malloc and causing assertion failures.

### DIFF
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -792,7 +792,7 @@ static ID3D11InputLayout* D3D11_INTERNAL_FetchBindingsInputLayout(
 	D3D11Renderer *renderer,
 	FNA3D_VertexBufferBinding *bindings,
 	int32_t numBindings,
-	int32_t *hash
+	uint32_t *hash
 ) {
 	int32_t numElements, bufsize, i, j, k, usage, index, attribLoc, bytecodeLength;
 	uint8_t attrUse[MOJOSHADER_USAGE_TOTAL][16];
@@ -807,12 +807,15 @@ static ID3D11InputLayout* D3D11_INTERNAL_FetchBindingsInputLayout(
 	/* We need the vertex shader... */
 	MOJOSHADER_d3d11GetBoundShaders(&vertexShader, &blah);
 
+	int32_t bindingsIndex;
+
 	/* Can we just reuse an existing input layout? */
 	result = (ID3D11InputLayout*) PackedVertexBufferBindingsArray_Fetch(
 		renderer->inputLayoutCache,
 		bindings,
 		numBindings,
 		vertexShader,
+		&bindingsIndex,
 		hash
 	);
 	if (result != NULL)
@@ -2031,7 +2034,8 @@ static void D3D11_ApplyVertexBufferBindings(
 	D3D11Renderer *renderer = (D3D11Renderer*) driverData;
 	D3D11Buffer *vertexBuffer;
 	ID3D11InputLayout *inputLayout;
-	int32_t i, stride, offset, hash;
+	int32_t i, stride, offset;
+	uint32_t hash;
 
 	if (!bindingsUpdated && !renderer->effectApplied)
 	{
@@ -4017,6 +4021,16 @@ static void D3D11_AddDisposeEffect(
 		renderer->currentPass = 0;
 		renderer->effectApplied = 1;
 	}
+
+	// invalidate all inputLayouts
+	for (int i = 0; i < renderer->inputLayoutCache.count; i += 1)
+	{
+		ID3D11InputLayout_Release(
+			(ID3D11InputLayout*)renderer->inputLayoutCache.elements[i].value
+		);
+	}
+	renderer->inputLayoutCache.count = 0;
+
 	MOJOSHADER_deleteEffect(effectData);
 	SDL_UnlockMutex(renderer->ctxLock);
 	SDL_free(effect);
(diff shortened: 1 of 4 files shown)


src/FNA3D_Driver_D3D11.c
```diff
@@ -807,12 +807,15 @@ static ID3D11InputLayout* D3D11_INTERNAL_FetchBindingsInputLayout(
 	/* We need the vertex shader... */
 	MOJOSHADER_d3d11GetBoundShaders(&vertexShader, &blah);
 
+	int32_t bindingsIndex;
```
flibitijibibo, line 810: Needs to be at the top of the function

src/FNA3D_Driver_D3D11.c
```diff
@@ -4017,6 +4021,16 @@ static void D3D11_AddDisposeEffect(
 		renderer->currentPass = 0;
 		renderer->effectApplied = 1;
 	}
+
+	// invalidate all inputLayouts
+	for (int i = 0; i < renderer->inputLayoutCache.count; i += 1)
```
flibitijibibo, line 4026: `int i` needs to be declared at the top of the function

src/FNA3D_Driver_D3D11.c
```diff
@@ -4017,6 +4021,16 @@ static void D3D11_AddDisposeEffect(
 		renderer->currentPass = 0;
 		renderer->effectApplied = 1;
 	}
+
+	// invalidate all inputLayouts
```
flibitijibibo, line 4025: `/* */` comments are required for C

src/FNA3D_Driver_Vulkan.c
```diff
@@ -7960,19 +7960,21 @@ static void VULKAN_ApplyVertexBufferBindings(
 	VulkanRenderer *renderer = (VulkanRenderer*) driverData;
 	MOJOSHADER_vkShader *vertexShader, *blah;
 	int32_t i, bindingsIndex;
+	uint32_t hash;
 	void* bindingsResult;
 	VulkanBuffer *vertexBuffer;
 	VulkanSubBuffer subbuf;
 	VkDeviceSize offset;
-
+
```
flibitijibibo, line 7968:
```suggestion

```

flibitijibibo (changes requested):
Mostly C compliance, otherwise I'll leave the rest of review to @TheSpydog...

src/FNA3D_Driver_D3D11.c
bartwe: updated

src/FNA3D_Driver_D3D11.c
bartwe: updated

src/FNA3D_Driver_D3D11.c
bartwe: updated

src/FNA3D_Driver_Vulkan.c
bartwe: updated

src/FNA3D_Driver_D3D11.c
```diff
@@ -807,12 +807,14 @@ static ID3D11InputLayout* D3D11_INTERNAL_FetchBindingsInputLayout(
 	/* We need the vertex shader... */
 	MOJOSHADER_d3d11GetBoundShaders(&vertexShader, &blah);
 
+
```
flibitijibibo, line 810:
```suggestion
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -4004,6 +4007,8 @@ static void D3D11_AddDisposeEffect(
 	FNA3D_Renderer *driverData,
 	FNA3D_Effect *effect
 ) {
+	int i;
```
flibitijibibo, line 4010:
```suggestion
	int32_t i;
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -4004,6 +4007,8 @@ static void D3D11_AddDisposeEffect(
...
+
```
flibitijibibo, line 4011:
```suggestion
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -4017,6 +4022,16 @@ static void D3D11_AddDisposeEffect(
 		renderer->currentPass = 0;
 		renderer->effectApplied = 1;
 	}
+
+	/* invalidate all inputLayouts */
+	for ( i = 0; i < renderer->inputLayoutCache.count; i += 1)
```
flibitijibibo, line 4027:
```suggestion
	for (i = 0; i < renderer->inputLayoutCache.count; i += 1)
```

flibitijibibo (changes requested):
All of these can be autocommitted I think. At a second glance it does seem pretty broad to just nuke every input layout ever vs only removing the entries with the relevant shaders, but I don't know how easy it is to remove entries in the middle, for example.

bartwe:
I avoided doing the filtering as it wasn't clear to me if the index was being stored by other parts of the system

flibitijibibo:
Latest revision's style is good. It's sounding like we need to treat input layout storage a bit differently; in particular it actually seems like the idea of linking input layouts to the vertex shader is unnecessary in the first place, given that A: Vulkan doesn't care about it at all, and B: D3D11 only uses the vertex shader for validation, nothing more:

https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11device-createinputlayout

>Once an input-layout object is created from a shader signature, the input-layout object can be reused with any other shader that has an identical input signature (semantics included). This can simplify the creation of input-layout objects when you are working with many shaders with identical inputs.

flibitijibibo:
Index/Hash portion has been applied:

https://github.com/FNA-XNA/FNA3D/commit/bbd8c758ea45bbe67f3efdf9779d3e1b6d4d8ead

Rejecting the input layout cache invalidator since we need a real solution to the deeper issue, which almost certainly affects Vulkan as well.

PR #53, opened 2020-09-22
WIP: Threaded Command Submission

thatcosmonaut:
This patch will handle Vulkan command preparation, submission, and presentation on a separate thread.


thatcosmonaut:
We are still hitting a brick wall on command encoding with this patch so I am going to take a different approach to building and submitting command buffers.

PR #59, opened 2020-10-23
Vulkan: Minor Allocator Tweaks

thatcosmonaut:
Games like SOR4 are crashing on low-VRAM devices due to high texture memory usage. 

Unfortunately it turns out that storing texture data on host memory is Hard, but this patch adds sub-allocators for the case where we use different memory types and makes a few minor tweaks to the allocation system.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -4869,8 +4869,8 @@ static uint8_t VULKAN_INTERNAL_FindAvailableTextureMemory(
 
 	if (requiredSize > allocator->nextAllocationSize)
 	{
-		/* allocate a page of required size aligned to 100MB increments */
-		allocator->nextAllocationSize =
+		/* allocate a page of required size aligned to ALLOCATION_SIZE increments */ allocator->nextAllocationSize
+		=
 			VULKAN_INTERNAL_NextHighestAlignment(requiredSize, TEXTURE_ALLOCATION_SIZE);
 	}
 
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -78,6 +78,7 @@ static uint32_t deviceExtensionCount = SDL_arraysize(deviceExtensionNames);
 #define MAX_UNIFORM_DESCRIPTOR_SETS 1024
 #define COMMAND_LIMIT 100
 #define TEXTURE_ALLOCATION_SIZE 100000000 /* 100MB */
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_Vulkan.c
```diff
@@ -7317,24 +7369,28 @@ static void VULKAN_DestroyDevice(FNA3D_Device *device)
 		NULL
 	);
 
-	for (i = 0; i < renderer->textureAllocator->textureBufferCount; i += 1)
+	for (i = 0; i < VK_MAX_MEMORY_TYPES; i++)
```
flibitijibibo, line 7372:
```suggestion
	for (i = 0; i < VK_MAX_MEMORY_TYPES; i += 1)
```

flibitijibibo (commented):
Will check this out in a bit!

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2094,6 +2102,12 @@ static uint8_t VULKAN_INTERNAL_FindMemoryType(
 			(memoryProperties.memoryTypes[i].propertyFlags & properties) == properties	)
 		{
 			*result = i;
+			FNA3D_LogInfo(
+				"Found suitable memory type; type index: %u, type filter: %u, properties: %u",
```
flibitijibibo, line 2106:
```suggestion
				"Found suitable memory type; type index: %u, type filter: %X, properties: %X",
```

I also ended up changing the LogError below to this:

```
FNA3D_LogWarn("Failed to find memory properties %X, filter %X\n", properties, typeFilter);
```

flibitijibibo (commented):
Also see suggestion from previous review

src/FNA3D_Driver_Vulkan.c
```diff
@@ -4869,22 +4867,27 @@ static uint8_t VULKAN_INTERNAL_FindAvailableTextureMemory(
...
-		allocationResult = VULKAN_INTERNAL_AllocateTextureMemory(
-			renderer,
-			memoryTypeIndex,
-			image,
-			requiredSize,
-			dedicatedRequirements.prefersDedicatedAllocation,
-			&textureBuffer
-		);
+		/* allocate a page of required size aligned to ALLOCATION_SIZE increments */ allocator->nextAllocationSize
+		=
```
flibitijibibo, line 4873: Looks like a newline went rogue...

```suggestion
		/* allocate a page of required size aligned to ALLOCATION_SIZE increments */
		allocator->nextAllocationSize =
```

PR #61, opened 2020-10-27
Vulkan: sort free texture memory regions by size

thatcosmonaut:
Instead of iterating over every texture memory page and free region when looking for available texture memory, we now maintain a list of free regions at the allocator level, sorted by size.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -947,23 +947,17 @@ typedef struct BufferMemoryWrapper
 
 typedef struct VulkanBuffer VulkanBuffer;
 
+typedef struct VulkanTextureBuffer VulkanTextureBuffer;
+
 /* Info about free regions of data within a VulkanTextureBuffer */
 typedef struct VulkanFreeTextureRegion
 {
+	VulkanTextureBuffer *buffer;
 	VkDeviceSize offset;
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_Vulkan.c
```diff
@@ -947,31 +947,44 @@ typedef struct BufferMemoryWrapper
...
+	 * from largest to smallest for efficient texture allocation
+	 */
+	VulkanFreeTextureRegion **sortedFreeRegions;
+	uint32_t sortedFreeRegionCount;
+	uint32_t sortedFreeRegionCapacity;
+} VulkanTextureSubAllocator;
+
 /* Manages memory for a single region of allocated texture memory */
-typedef struct VulkanTextureBuffer
+struct VulkanTextureBuffer
```
flibitijibibo, line 978: Should this still be a typedef?

flibitijibibo (commented):
Not much to see here, will check performance first thing in the morning.

src/FNA3D_Driver_Vulkan.c
thatcosmonaut: It has to be forward-declared because of mutual reference.

flibitijibibo: Woops, totally missed that, my mistake!

PR #63, opened 2020-10-29
Vulkan: Unify buffer and texture allocation systems

thatcosmonaut:
This patch lets us leverage our free memory region strategy that we used for texture memory to allocate buffer space. Now we use the same memory allocation system for both types of allocations. We can now properly reclaim buffer space when buffers are destroyed. This also means that we get proper memory alignment for buffers, and we no longer have to crash if we run out of buffer space, among other advantages.


flibitijibibo:
I wasn't able to reproduce any issues this since last week's test... so either I had a bad version or it was just a fluke. Either way, it's probably better to have this than the current None situation, so I'll squash this in.

PR #66, opened 2020-11-06
Copy NONE buffer data instead of flush and stall

thatcosmonaut:
Changes SetBufferData with NONE option to use a buffer copy command on a bound buffer instead of flushing commands and stalling. This is what DXVK does for D3D9. Additionally, our internal list of bound buffers now tracks VulkanSubBuffers instead of VulkanBuffers, since these map directly to used VkBuffers.


thatcosmonaut:
There are some unforeseen structural problems with this PR. Will fix these and open a new PR.

PR #68, opened 2020-11-07
Vulkan: fix null free on freeing dedicated resource

thatcosmonaut:
The sole free region created by an allocation is consumed entirely by a dedicated alloc, so these frees crash the program.


flibitijibibo:
What I get for trying to amend commits on a phone... will fix and force push later today.

PR #74, opened 2020-11-20
Vulkan: More well-behaved present behavior

thatcosmonaut:
NVIDIA throws KHR_ERROR on minimize and alt-tab, so our previous swapchain behavior was crashing. We were also bailing hard out of all draw work if we failed to acquire a swapchain image. What we now do is when KHR_ERROR comes back from AcquireSwapchain, we cancel presentation, but still submit the queued work and rotate references. RTs can render and objects get properly cleaned up even if we don't present. Then we attempt to recreate the swapchain after work is submitted.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -4185,6 +4185,7 @@ static void VULKAN_INTERNAL_SubmitCommands(
 	VkSubmitInfo submitInfo;
 	uint32_t i, j;
 	VkResult result, presentResult = VK_SUCCESS;
+	uint8_t acquireSuccess = 0;
 	uint32_t swapChainImageIndex;
 
 	FNA3D_Rect *sourceRectangle, *destinationRectangle;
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_Vulkan.c
```diff
@@ -4466,21 +4423,25 @@ static void VULKAN_INTERNAL_SubmitCommands(
...
-	/* Now that commands are totally done, check for present errors */
+	/* Now that commands are totally done, check if we need new swapchain */
 	if (present)
 	{
-		if (	presentResult == VK_ERROR_OUT_OF_DATE_KHR ||
-			presentResult == VK_SUBOPTIMAL_KHR	)
+		if (acquireResult == VK_ERROR_OUT_OF_DATE_KHR ||
+			acquireResult == VK_SUBOPTIMAL_KHR || 
+			presentResult == VK_ERROR_OUT_OF_DATE_KHR ||
+			presentResult == VK_SUBOPTIMAL_KHR)
```
flibitijibibo, line 4432:
```suggestion
		if (	acquireResult == VK_ERROR_OUT_OF_DATE_KHR ||
			acquireResult == VK_SUBOPTIMAL_KHR || 
			presentResult == VK_ERROR_OUT_OF_DATE_KHR ||
			presentResult == VK_SUBOPTIMAL_KHR	)
```

PR #76, opened 2020-12-03
Vulkan: Fix multisample and depth stencil RT texture leaks

thatcosmonaut:
This patch fixes `VULKAN_AddDisposeRenderbuffer` not disposing textures created when generating a render target.


jeffphilp:
I think 1e4b66b commit combined with 491a26e can lead to a segmentation fault, as the disposal code for render buffers `VULKAN_INTERNAL_DestroyRenderBuffer` invoked from `VULKAN_INTERNAL_PerformDeferredDestroys` destroys the textures but it wasn't originally getting invoked due to the typo fixed in 1e4b66b.  So now with 491a26e also scheduling a destory it ends up happening twice.

thatcosmonaut:
@jeffphilp You are absolutely right! Nice catch. I'll fix this.

PR #77, opened 2020-12-03
Add cmake option to disable D3D11 backend.

madewokherd:
The D3D11 backend doesn't currently work on Wine for multiple reasons, the most important of which is that Wine doesn't have an HLSL shader compiler. I'd like to update Wine Mono's shipped FNA to a version using FNA3D, and disabling the D3D11 backend is the simplest way to get it working.

I don't think HLSL could be used on any of the other backends, so the option disables that as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ project(FNA3D C)
 
 # Options
 option(BUILD_SHARED_LIBS "Build shared library" ON)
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Thought that might come up at some point... I'm hoping to bump Vulkan/MoltenVK to the very top next year so hopefully this will be made partially obsolete once that's done.

PR #79, opened 2020-12-17
cmake: use pkgconfig to find sdl2

bl4ckb0ne:
This fixes the build of `FNA3D` on alpine linux. Somehow the SDL2 package on alpine has a dependency to directfb for one project, causing the following error

```
In file included from /home/simon/dev/FNA3D/src/FNA3D_Driver_OpenGL.c:33:
/usr/include/SDL2/SDL_syswm.h:80:10: fatal error: directfb.h: No such file or directory
   80 | #include <directfb.h>
      |          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/FNA3D.dir/build.make:108: CMakeFiles/FNA3D.dir/src/FNA3D_Driver_OpenGL.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from /home/simon/dev/FNA3D/src/FNA3D_Driver_Vulkan.c:36:
/usr/include/SDL2/SDL_syswm.h:80:10: fatal error: directfb.h: No such file or directory
   80 | #include <directfb.h>
      |          ^~~~~~~~~~~~
compilation terminated.
```

Using `pkgconfig` to get the SDL2 library and include paths gives the correct path to directfb.

```
$ sdl2-config --cflags
-I/usr/include/SDL2 -I/usr/include/directfb -D_REENTRANT -D_REENTRANT
```

I will test the patch tomorrow on a glibc distro to validate it. I don't own any Windows or MacOS computer, do you know how can I test on these platforms?


flibitijibibo:
This new block makes sense, but we shouldn't delete what's already there - if we can introduce the pkg-config path in a way that doesn't invade on the existing paths that others depend on, this should be very safe to merge.

bl4ckb0ne:
Hmmm, what about using pkg-config only on linux, and let the other platforms use the target from `find_package`?

flibitijibibo:
The way it works now, we might be able to expand that last else block to use pkg-config if neither SDL2_LIBRARIES nor SDL2_INCLUDE_DIRS are defined - we sorta just pray that at least one is specified, but I think if we know for sure that neither exists we can fall back to pkg-config.

bl4ckb0ne:
My system generates the build files with using the `SDL2::SDL2` target, but it misses the directfb dependency because the `FindSDL2.cmake` file searches only for the library and the include file.

flibitijibibo:
Ah - in that case I would file an issue on the SDL Bugzilla, we have other issues with the CMake file anyway so more eyes on that would be great.

bl4ckb0ne:
Sure, could you link the the other issues please? I can work on that.

flibitijibibo:
This is the one I was looking at, it's more in MinGW territory though: https://bugzilla.libsdl.org/show_bug.cgi?id=5115

bl4ckb0ne:
Can't do much for mingw, sorry. I'll close that request and work on a fix for SDL2 directly.

PR #80, opened 2020-12-18
cmake: use pkgconfig to find sdl2 on unix platforms

bl4ckb0ne:
Hey, its me again sorry. Turns out the `SDL2Targets.cmake` is a generated file and I couldn't find a way to add the path to directfb's headers in the generation step.

I reworked the previous patch to use pkgconfig on unix platforms only. To my knowledge all linux/bsd should support it. I'll try to find a way to test it on windows/macos tomorrow.


flibitijibibo:
Like the previous patch this is still far too invasive on other systems - I strongly recommend filing this on the SDL Bugzilla and looking at ways to add this to their .cmake generator. It's very likely that this affects other backends too, it just happens to be that directfb is the one that's been caught in this instance. If they set the flags correctly for sdl2-config they should also do it for the SDL CMake files.

flibitijibibo:
Found some historical bugs exhibiting this issue, so this is for sure an SDL issue:

https://bugs.gentoo.org/503130

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=669363

http://lists.busybox.net/pipermail/buildroot/2018-January/211971.html

bl4ckb0ne:
Bug opened https://bugzilla.libsdl.org/show_bug.cgi?id=5403

PR #85, opened 2021-01-06
Add support for sRGB textures, render targets and backbuffers

kg:
Needs an FNA change to expose the new enum value but that's like two lines

### DIFF
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -227,6 +227,7 @@ typedef struct D3D11Renderer /* Cast FNA3D_Renderer* to this! */
 	uint8_t debugMode;
 	uint32_t supportsDxt1;
 	uint32_t supportsS3tc;
+	uint8_t supportsSRGBRenderTarget;
 	int32_t maxMultiSampleCount;
 	D3D_FEATURE_LEVEL featureLevel;
 
@@ -307,6 +308,8 @@ static DXGI_FORMAT XNAToD3D_TextureFormat[] =
 	DXGI_FORMAT_R16G16B16A16_FLOAT,	/* SurfaceFormat.HalfVector4 */
 	DXGI_FORMAT_R16G16B16A16_FLOAT,	/* SurfaceFormat.HdrBlendable */
 	DXGI_FORMAT_B8G8R8A8_UNORM,	/* SurfaceFormat.ColorBgraEXT */
+	DXGI_FORMAT_R8G8B8A8_UNORM_SRGB,/* SurfaceFormat.ColorSrgbEXT */
+	DXGI_FORMAT_BC3_UNORM_SRGB,	/* SurfaceFormat.Dxt5SrgbEXT */
 };
 
 static DXGI_FORMAT XNAToD3D_DepthFormat[] =
@@ -2495,6 +2498,16 @@ static void D3D11_INTERNAL_CreateBackbuffer(
 	useFauxBackbuffer = (	useFauxBackbuffer ||
 				parameters->multiSampleCount > 0	);
 
+	FNA3D_SurfaceFormat actualFnaFormat = parameters->backBufferFormat;
+	DXGI_FORMAT actualDxgiFormat = XNAToD3D_TextureFormat[actualFnaFormat];
+
+	if ((actualDxgiFormat == DXGI_FORMAT_R8G8B8A8_UNORM_SRGB) && !renderer->supportsSRGBRenderTarget)
+	{
+		FNA3D_LogWarn("Could not create an SRGB swapchain because this renderer does not support it. Silently falling back to UNORM.");
+		actualFnaFormat = FNA3D_SURFACEFORMAT_COLOR;
+		actualDxgiFormat = DXGI_FORMAT_R8G8B8A8_UNORM;
+	}
+
 	if (useFauxBackbuffer)
 	{
 		if (	renderer->backbuffer == NULL ||
@@ -2524,7 +2537,7 @@ static void D3D11_INTERNAL_CreateBackbuffer(
 		colorBufferDesc.Height = renderer->backbuffer->height;
 		colorBufferDesc.MipLevels = 1;
 		colorBufferDesc.ArraySize = 1;
-		colorBufferDesc.Format = XNAToD3D_TextureFormat[renderer->backbuffer->d3d11.surfaceFormat];
+		colorBufferDesc.Format = actualDxgiFormat;
 		colorBufferDesc.SampleDesc.Count = (
 			renderer->backbuffer->multiSampleCount > 1 ?
 				renderer->backbuffer->multiSampleCount :
@@ -2573,7 +2586,7 @@ static void D3D11_INTERNAL_CreateBackbuffer(
 			colorBufferDesc.Height = renderer->backbuffer->height;
 			colorBufferDesc.MipLevels = 1;
 			colorBufferDesc.ArraySize = 1;
-			colorBufferDesc.Format = XNAToD3D_TextureFormat[renderer->backbuffer->d3d11.surfaceFormat];
+			colorBufferDesc.Format = actualDxgiFormat;
 			colorBufferDesc.SampleDesc.Count = 1;
 			colorBufferDesc.SampleDesc.Quality = 0;
 			colorBufferDesc.Usage = D3D11_USAGE_DEFAULT;
@@ -2625,6 +2638,7 @@ static void D3D11_INTERNAL_CreateBackbuffer(
 		renderer->backbuffer->width = parameters->backBufferWidth;
 		renderer->backbuffer->height = parameters->backBufferHeight;
 		renderer->backbuffer->depthFormat = parameters->depthStencilFormat;
+		renderer->backbuffer->d3d11.surfaceFormat = parameters->backBufferFormat;
 		renderer->backbuffer->multiSampleCount = 0;
 	}
 
@@ -2689,7 +2703,10 @@ static void D3D11_INTERNAL_CreateBackbuffer(
 	}
 
 	/* Create a render target view for the swapchain */
-	swapchainViewDesc.Format = DXGI_FORMAT_R8G8B8A8_UNORM;
+	DXGI_FORMAT swapchainFormat = (actualDxgiFormat == DXGI_FORMAT_R8G8B8A8_UNORM_SRGB)
+		? DXGI_FORMAT_R8G8B8A8_UNORM_SRGB
+		: DXGI_FORMAT_R8G8B8A8_UNORM;
+	swapchainViewDesc.Format = swapchainFormat;
 	swapchainViewDesc.ViewDimension = D3D11_RTV_DIMENSION_TEXTURE2D;
 	swapchainViewDesc.Texture2D.MipSlice = 0;
 
@@ -3230,13 +3247,12 @@ static void D3D11_SetTextureData2D(
 	D3D11Texture *d3dTexture = (D3D11Texture*) texture;
 	D3D11_BOX dstBox;
 
-	/* DXT formats require w and h to be multiples of 4 */
-	if (	d3dTexture->format == FNA3D_SURFACEFORMAT_DXT1 ||
-		d3dTexture->format == FNA3D_SURFACEFORMAT_DXT3 ||
-		d3dTexture->format == FNA3D_SURFACEFORMAT_DXT5	)
+	int32_t blockSize = Texture_GetBlockSize(d3dTexture->format);
+
+	if (blockSize > 1)
 	{
-		w = (w + 3) & ~3;
-		h = (h + 3) & ~3;
+		w = (w + blockSize - 1) & ~(blockSize - 1);
+		h = (h + blockSize - 1) & ~(blockSize - 1);
 	}
 
 	dstBox.left = x;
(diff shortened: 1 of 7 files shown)


src/FNA3D_Driver_D3D11.c
```diff
@@ -2519,7 +2520,11 @@ static void D3D11_INTERNAL_CreateFramebuffer(
 	}
 
 	/* Create a render target view for the swapchain */
-	swapchainViewDesc.Format = DXGI_FORMAT_R8G8B8A8_UNORM;
+	
+	DXGI_FORMAT swapchainFormat = colorBufferDesc.Format == DXGI_FORMAT_R8G8B8A8_UNORM_SRGB
```
flibitijibibo, line 2524:
```suggestion
	DXGI_FORMAT swapchainFormat = (colorBufferDesc.Format == DXGI_FORMAT_R8G8B8A8_UNORM_SRGB)
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -2544,11 +2548,20 @@ static void OPENGL_SetRenderTargets(
 				renderer->realBackbufferFBO
 		);
 		renderer->renderTargetBound = 0;
+		/* The driver is able and willing to provide us a sRGB backbuffer even if we don't ask for one,
+		 * so we need to disable FRAMEBUFFER_SRGB if we don't actually want sRGB blending/sampling */
+		if (renderer->backbuffer->isSrgb)
+			renderer->glEnable(GL_FRAMEBUFFER_SRGB_EXT);
+		else
+			renderer->glDisable(GL_FRAMEBUFFER_SRGB_EXT);
```
flibitijibibo, line 2556:
```suggestion
		/* The driver is able and willing to provide us a sRGB backbuffer even if we don't ask for one,
		 * so we need to disable FRAMEBUFFER_SRGB if we don't actually want sRGB blending/sampling
		 */
		if (renderer->backbuffer->isSrgb)
		{
			renderer->glEnable(GL_FRAMEBUFFER_SRGB_EXT);
		}
		else
		{
			renderer->glDisable(GL_FRAMEBUFFER_SRGB_EXT);
		}
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -2544,11 +2548,20 @@ static void OPENGL_SetRenderTargets(
...
 		return;
 	}
 	else
 	{
 		BindFramebuffer(renderer, renderer->targetFramebuffer);
+		/* unlike the backbuffer, our format will be sRGB or non-sRGB as we expect so we can leave
+		 * this flag on universally and the blending will be configured based on the format */
+		renderer->glEnable(GL_FRAMEBUFFER_SRGB_EXT);
```
flibitijibibo, line 2564:
```suggestion
		/* Unlike the backbuffer, our format will be sRGB or non-sRGB as we expect so we can leave
		 * this flag on universally and the blending will be configured based on the format
		 */
		renderer->glEnable(GL_FRAMEBUFFER_SRGB_EXT);
```

flibitijibibo (changes requested):
Mostly style, though for OpenGL...

- We need a state shadowing toggle for `GL_FRAMEBUFFER_SRGB_EXT`.
- We also need an extension check for `GL_EXT_framebuffer_sRGB`, can rip off the S3TC check

Also, until we finish Vulkan this will need a Metal implementation as well.

kg:
Can you explain what you mean by state shadowing toggle?

flibitijibibo:
For example:

```
uint8_t srgbEnabled; /* Declared in the renderer */

uint8_t newSRGB; /* Determined in the function */
if (newSGRB != renderer->srgbEnabled)
{
    renderer->srgbEnabled = newSRGB;
    ToggleGLState(renderer, GL_FRAMEBUFFER_SRGB_EXT, renderer->srgbEnabled);
}
```

flibitijibibo:
Converted to a draft for now - D3D11 and OpenGL have a lot of work before being considered ready. The Metal enum is thankfully just the one added enum to the array lookup so that one should be easy.

kg:
Here for posterity: This PR turns out to be not so useful unless we also add SRGB versions of the block texture formats (DXT1/3/5). Going to figure out how hard that is to add...

kg:
Added DXT5 SRGB variant, not sure it's worth adding 1/3 (I can do it if you want). Tested that in Vulkan and D3D11 but it seems to be wonky in OpenGL...

flibitijibibo:
We can skip 1/3 I think.

D3D and Vulkan look a whole hell of a lot better now, OpenGL I'm still a bit worried about the backbuffer assumption (and it sounds like GL needs more stability fixes anyhow).

As a quick reminder we still need to incorporate buildfixes for Metal, it should just be adding a whopping 2 enums but maybe the block size stuff has changed that.

kg:
I think the GL issues are actually just due to my use of block textures, because my block texture uploads are producing validator warnings in Vulkan. Hopefully once we get the validator warnings figured out it will indicate what I'm doing wrong and GL will turn out to be working... if it's not I'll keep digging in there.

flibitijibibo:
Metal has been cut, so this should be rebased.

kg:
I rebased + squished and then cleaned up the indentation after using .editorconfig to set my indent level to 8. Will do some testing locally to see if anything is weird.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2154,7 +2160,7 @@ static uint8_t VULKAN_INTERNAL_ChooseSwapSurfaceFormat(
 	for (i = 0; i < availableFormatsLength; i += 1)
 	{
 		if (	availableFormats[i].format == desiredFormat &&
-			availableFormats[i].colorSpace == VK_COLOR_SPACE_SRGB_NONLINEAR_KHR	)
+			availableFormats[i].colorSpace == VK_COLOR_SPACE_SRGB_NONLINEAR_KHR)
```
flibitijibibo, line 2163: Tab got lost

flibitijibibo (approved):
Looks good, long as it passes everyone's tests this'll go in for 22.01.

src/FNA3D_Driver_Vulkan.c
kg: Tab looked like an error to me, what's the rule for it being there so I know where else to put it?

kg:
The OpenGL implementation broke since the last time I messed with it, so I need to figure out what's going on there. I got an INVALID_VALUE error once in VS that doesn't happen under Renderdoc, and it seems like the relevant sRGB flags get flipped at random while things are running. :(

D3D11 and Vulkan seem fine, though I'm having issues with one of my tests that have me a bit confused about what's going on with rendertargets...

kg:
Anyone wanting to test this with FNA proper can use https://github.com/sq/FNA/tree/srgb. The rendertarget issues were just me forgetting to change the landmine if-statement in Texture2D that tricks me every time I mess with sRGB.

kg:
The INVALID_VALUE in GL was from me using block textures, it doesn't happen if I switch them off and those work right in D3D11 and Vulkan. Something about what's expected for DXT block sizes in GL is weird - if someone else can run tests there that would be appreciated. I'm still seeing some weirdness where the framebuffer sRGB flag seems to get ignored sometimes in OpenGL and then it suddenly applies and sticks around.

src/FNA3D_Driver_Vulkan.c
flibitijibibo: It's a tab opposite to the one at the start of the parens, the one after the `if (`.

Now that I'm not on my phone...

```suggestion
			availableFormats[i].colorSpace == VK_COLOR_SPACE_SRGB_NONLINEAR_KHR	)
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -358,7 +364,9 @@ static int32_t XNAToGL_TextureInternalFormat[] =
 	GL_RG16F,				/* SurfaceFormat.HalfVector2 */
 	GL_RGBA16F,				/* SurfaceFormat.HalfVector4 */
 	GL_RGBA16F,				/* SurfaceFormat.HdrBlendable */
-	GL_RGBA8				/* SurfaceFormat.ColorBgraEXT */
+	GL_RGBA8,				/* SurfaceFormat.ColorBgraEXT */
+	GL_SRGB_ALPHA_EXT,			/* SurfaceFormat.ColorSrgbEXT */
+	GL_COMPRESSED_RGBA_S3TC_DXT5_EXT,	/* SurfaceFormat.Dxt5SrgbEXT */
```
flibitijibibo, line 369: Looking around this file, I'm wondering if this doesn't work because it's no different from the regular DXT5 values? Is there supposed to be a DXT5 SRGB enum somewhere?

src/FNA3D_Driver_D3D11.c
```diff
@@ -2625,6 +2638,7 @@ static void D3D11_INTERNAL_CreateBackbuffer(
 		renderer->backbuffer->width = parameters->backBufferWidth;
 		renderer->backbuffer->height = parameters->backBufferHeight;
 		renderer->backbuffer->depthFormat = parameters->depthStencilFormat;
+		renderer->backbuffer->d3d11.surfaceFormat = parameters->backBufferFormat;
```
flibitijibibo, line 2641: Should this be actualFnaFormat?

src/FNA3D_Driver_D3D11.c
```diff
@@ -2689,7 +2703,10 @@ static void D3D11_INTERNAL_CreateBackbuffer(
 	}
 
 	/* Create a render target view for the swapchain */
-	swapchainViewDesc.Format = DXGI_FORMAT_R8G8B8A8_UNORM;
+	DXGI_FORMAT swapchainFormat = (actualDxgiFormat == DXGI_FORMAT_R8G8B8A8_UNORM_SRGB)
```
flibitijibibo, line 2706: Can remove this local variable, plus it probably breaks on some C compilers

flibitijibibo (changes requested):
Taking another shot at this...

Something else of note: For OpenGL you will probably want to check `SDL_GL_FRAMEBUFFER_SRGB_CAPABLE` as an attribute, to ensure the backbuffer is capable of supporting the format.

src/FNA3D_Driver_D3D11.c
kg: Oh are ternary expressions not available in classic C?

flibitijibibo: They're available, but variables mid-function aren't supported by older MSVC releases. I think you can just assign this directly to the struct member right below.

src/FNA3D_Driver_OpenGL.c
```diff
@@ -1110,6 +1120,15 @@ static inline void ToggleGLState(
 	}
 }
 
+static inline void ApplySRGBFlag(OpenGLRenderer *renderer, uint8_t state)
+{
+	if (state == renderer->srgbEnabled)
+		return;
```
flibitijibibo, line 1126:
```suggestion
	if (state == renderer->srgbEnabled)
	{
		return;
	}
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -3142,8 +3170,12 @@ static void OPENGL_INTERNAL_CreateBackbuffer(
 		renderer->backbuffer->width = parameters->backBufferWidth;
 		renderer->backbuffer->height = parameters->backBufferHeight;
 		renderer->backbuffer->depthFormat = renderer->windowDepthFormat;
+		renderer->backbuffer->isSrgb = parameters->backBufferFormat == FNA3D_SURFACEFORMAT_COLORSRGB_EXT;
 		renderer->backbuffer->multiSampleCount = 0;
 	}
+
+	if (renderer->backbuffer)
+		ApplySRGBFlag(renderer, renderer->backbuffer->isSrgb);
```
flibitijibibo, line 3178:
```suggestion
	if (renderer->backbuffer)
	{
		ApplySRGBFlag(renderer, renderer->backbuffer->isSrgb);
	}
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -2552,13 +2572,16 @@ static void OPENGL_SetRenderTargets(
 		renderer->renderTargetBound = 1;
 	}
 
+	uint8_t isSrgb = 0;
```
flibitijibibo, line 2575: Variable mid-function

src/FNA3D_Driver_D3D11.c
```diff
@@ -2495,6 +2498,16 @@ static void D3D11_INTERNAL_CreateBackbuffer(
 	useFauxBackbuffer = (	useFauxBackbuffer ||
 				parameters->multiSampleCount > 0	);
 
+	FNA3D_SurfaceFormat actualFnaFormat = parameters->backBufferFormat;
+	DXGI_FORMAT actualDxgiFormat = XNAToD3D_TextureFormat[actualFnaFormat];
```
flibitijibibo, line 2502: Variables mid-function

src/FNA3D_Driver_Vulkan.c
```diff
@@ -10431,14 +10441,10 @@ static void VULKAN_INTERNAL_SetTextureData(
...
-	/* DXT texture buffers must be at least 4x4 */
-	if (	texture->colorFormat == FNA3D_SURFACEFORMAT_DXT1 ||
-		texture->colorFormat == FNA3D_SURFACEFORMAT_DXT3 ||
-		texture->colorFormat == FNA3D_SURFACEFORMAT_DXT5	)
-	{
-		bufferRowLength = SDL_max(4, w);
-		bufferImageHeight = SDL_max(4, h);
-	}
+	/* Block compressed texture buffers must be at least 1 block in width and height */
+	int32_t blockSize = Texture_GetBlockSize(texture->colorFormat);
```
flibitijibibo, line 10445: Variable mid-function

flibitijibibo (changes requested):
Final style pass, will merge when completed

kg:
OK, cleanup should be done

PR #86, opened 2021-01-08
Add GetSysRendererEXT and CreateSysTextureEXT

thatcosmonaut:
FNA users have often requested things like a modern shader system, and other graphics features. However, because FNA3D is primarily a preservation project, we don't want to provide complex rendering extensions that could compromise the stability of the library. This led to me implementing an entirely new graphics API, but the problem is that users might want to take advantage of this API without needing to rewrite their entire application.

My proposal is to expose a minimum amount of information that will allow FNA3D to export its rendering context and use an externally-created sampler view. This will allow the user to render using a separate API, while FNA3D composites the render into the frame and presents. 

This patch does the following:

1) Allows for the FNA3D-initialized rendering context to be read. This allows another library to piggyback on FNA3D's rendering context and share resources.

2) Allows a special type of sampler-only texture to be created by handing off handle info. For Vulkan, this is a VkImageView, for D3D the equivalent is a ShaderResourceView, for example.

This provides a minimally-invasive method for the user to integrate a separate graphics system into their FNA application.

I have a Vulkan proof-of-concept of this process working with FNA3D and Refresh: https://github.com/thatcosmonaut/RefreshFNA3DTest

### DIFF
--- a/include/FNA3D.h
+++ b/include/FNA3D.h
@@ -44,6 +44,19 @@
 #endif /* __GNUC__ */
 #endif /* FNA3DNAMELESS */
 
+#define VK_DEFINE_HANDLE(object) typedef struct object##_T* object;
+
+#if defined(__LP64__) || defined(_WIN64) || defined(__x86_64__) || defined(_M_X64) || defined(__ia64) || defined (_M_IA64) || defined(__aarch64__) || defined(__powerpc64__)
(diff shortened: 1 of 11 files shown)


thatcosmonaut:
Upon suggestion, this now uses an SDL_SysWM-style union struct to acquire rendering context in a backend-generic way.

PR #91, opened 2021-01-20
Vulkan: save and load pipeline cache

thatcosmonaut:
We were initializing a pipeline cache, but it never had any data! Now we save the pipeline cache data to a file in DestroyDevice and load it in CreateDevice.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -109,6 +109,8 @@ static const uint32_t deviceExtensionCount = SDL_arraysize(deviceExtensionNames)
 #define NULL_SAMPLER (VkSampler) 0
 #define NULL_RENDER_PASS (VkRenderPass) 0
 
+#define PIPELINE_CACHE_FILE_NAME "vk_pipeline.cache"
+
 #define IDENTITY_SWIZZLE \
 { \
 	VK_COMPONENT_SWIZZLE_IDENTITY, \
@@ -7345,6 +7347,8 @@ static void VULKAN_DestroyDevice(FNA3D_Device *device)
 	VulkanMemorySubAllocator *allocator;
 	uint32_t i, j, k;
 	VkResult waitResult;
+	size_t pipelineCacheSize;
+	uint8_t *pipelineCacheData;
 
 	VULKAN_INTERNAL_FlushCommands(renderer, 1);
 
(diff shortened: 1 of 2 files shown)


src/FNA3D_Driver_Vulkan.c
```diff
@@ -7355,6 +7361,36 @@ static void VULKAN_DestroyDevice(FNA3D_Device *device)
...
+		&pipelineCacheSize,
+		pipelineCacheData
+	);
+
+	pipelineCacheFileName = SDL_GetHint("FNA3D_VULKAN_PIPELINE_CACHE_FILE_NAME");
+	if (pipelineCacheFileName == NULL)
+	{
+		pipelineCacheFileName = DEFAULT_PIPELINE_CACHE_FILE_NAME;
+	}
+	pipelineCacheFile = SDL_RWFromFile(pipelineCacheFileName, "wb");
```
flibitijibibo, line 7387: Needs error handling and a LogWarn when we're unable to save the file.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -10570,14 +10609,40 @@ static FNA3D_Device* VULKAN_CreateDevice(
...
 	pipelineCacheCreateInfo.flags = 0;
-	pipelineCacheCreateInfo.initialDataSize = 0;
-	pipelineCacheCreateInfo.pInitialData = NULL;
+
+	pipelineCacheFileName = SDL_GetHint("FNA3D_VULKAN_PIPELINE_CACHE_FILE_NAME");
+	if (pipelineCacheFileName == NULL)
+	{
+		pipelineCacheFileName = DEFAULT_PIPELINE_CACHE_FILE_NAME;
+	}
+	pipelineCacheFile = SDL_RWFromFile(pipelineCacheFileName, "rb");
```
flibitijibibo, line 10618: Can be simplified with SDL_LoadFile

src/FNA3D_Driver_Vulkan.c
```diff
@@ -109,6 +109,8 @@ static const uint32_t deviceExtensionCount = SDL_arraysize(deviceExtensionNames)
 #define NULL_SAMPLER (VkSampler) 0
 #define NULL_RENDER_PASS (VkRenderPass) 0
 
+#define DEFAULT_PIPELINE_CACHE_FILE_NAME "vk_pipeline.cache"
```
flibitijibibo, line 112: This should probably be more descriptive, like FNA3D_Vulkan_PipelineCache.blob, so developers know where the new file came from.

flibitijibibo (changes requested):
Seems fine, just needs some minor hardening.

PR #94, opened 2021-01-26
Vulkan: textures fall back to host memory if out of device memory

thatcosmonaut:
Right now we have absurd device texture memory requirements because we crash on failing to allocate device local memory. This patch allows the renderer to fall back on host memory for most allocation cases. The exceptions are render targets, which should always be on device memory, and depth formats or BC* compression formats, which must be stored in optimal tiling and therefore cannot be stored in host memory.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -5560,7 +5560,7 @@ static uint8_t VULKAN_INTERNAL_CreateTexture(
 			return 0;
 		}
 
-		FNA3D_LogInfo("Out of device local memory, falling back to host memory");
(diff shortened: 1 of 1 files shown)


DadSchoorse:
@thatcosmonaut @flibitijibibo Just for your information: it's possible have optimal images in normal host memory on amd. Nvidia has special memory types that use host memory which isn't host visible `VkPhysicalDeviceMemoryProperties::propertyFlags == 0` (memory types 0 - 6 on my nvidia card). So afaik it should always be possible to fallback to host memory with optimal layout.

thatcosmonaut:
@DadSchoorse Looks like you're right, thanks for the tip-off.

PR #95, opened 2021-01-28
Vulkan: remove host-visible and linear tiling requirement from textures

thatcosmonaut:
Apparently we can just allocate textures on CPU memory as optimal if we remove the HOST_VISIBLE requirement, which we never actually need for textures, so that's a nice code complexity reduction and performance boost. It also means that we can allocate compressed textures to host memory.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -1551,37 +1551,6 @@ static inline uint8_t DepthFormatContainsStencil(VkFormat format)
 	}
 }
 
-static inline uint8_t ImageFormatCanBeCPUAllocated(VkFormat format)
-{
-	switch (format)
-	{
-		case VK_FORMAT_D16_UNORM:
-		case VK_FORMAT_D16_UNORM_S8_UINT:
-		case VK_FORMAT_D32_SFLOAT:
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Easy enough... thanks to @DadSchoorse for the tip!

src/FNA3D_Driver_Vulkan.c
```diff
@@ -3160,9 +3129,7 @@ static uint8_t VULKAN_INTERNAL_FindAvailableTextureMemory(
 
 	if (cpuAllocation)
 	{
-		memoryPropertyFlags =
-			VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT |
-			VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+		memoryPropertyFlags = VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
```
DadSchoorse, line 3132: This should be dropped as well, currently you won't find the memory host type on nvidia because its flags is `0`.

thatcosmonaut: Is it legal to allocate a texture on non-coherent memory?

thatcosmonaut: Well I just tested it and it works, so that answers that question!

PR #110, opened 2021-04-08
Vulkan Defragmenter

thatcosmonaut:
It's finally here! 

Every frame after commands are submitted, the system checks to see if any regions of memory are fragmented. If they are, we allocate a new block and queue up GPU-to-GPU copy commands of the resources in the fragmented block, which are added to a special command buffer that is submitted immediately. This way the copies are performed while the next frame is being built. The copies are performed sequentially into the new block, leaving no gaps. Once the next frame is submitted, we iterate over the resources that were copied and destroy them. If a block of memory is completely freed, we deallocate it.

All render targets are now allocated to dedicated memory blocks to prevent expensive copies and having to destroy and recreate a huge chain of resource dependencies.

We also now mark off partitions to include alignment offset regions to avoid tiny free regions created by alignments.

This is a very complex system and it's going to need a lot of testing, but I'm confident that this approach will reduce memory usage significantly.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -1042,6 +1042,7 @@ static VulkanTexture NullTexture =
 
 struct VulkanSubBuffer
 {
+	VulkanBuffer *parent;
 	VulkanMemoryUsedRegion *usedRegion;
 	VkBuffer buffer;
 	VulkanResourceAccessType resourceAccessType;
@@ -1808,6 +1809,11 @@ static void VULKAN_INTERNAL_ImageMemoryBarrier(
 	uint8_t discardContents,
 	VkImage image,
 	VulkanResourceAccessType *resourceAccessType
+);
+
+static void VULKAN_INTERNAL_RemoveBuffer(
+	FNA3D_Renderer *driverData,
+	FNA3D_Buffer *buffer
 );
 
 static void VULKAN_INTERNAL_RecreateSwapchain(VulkanRenderer *renderer, uint8_t flush);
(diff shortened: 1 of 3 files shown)


src/FNA3D_Driver_Vulkan.c
```diff
@@ -94,6 +94,7 @@ static const uint32_t deviceExtensionCount = SDL_arraysize(deviceExtensionNames)
 #define FAST_TEXTURE_STAGING_SIZE 64000000 /* 64MB */
 #define STARTING_SLOW_TEXTURE_STAGING_SIZE 16000000 /* 16MB */
 #define MAX_SLOW_TEXTURE_STAGING_SIZE 256000000 /* 256MB */
+#define DEVICE_LOCAL_HEAP_USAGE_FACTOR 0.8
```
kg, line 97: This might need to be a value you can set externally (via an env var or whatever)

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2490,6 +2546,22 @@ static uint8_t VULKAN_INTERNAL_DeterminePhysicalDevice(
 		&renderer->memoryProperties
 	);
 
+	deviceLocalHeapSize = 0;
+	for (i = 0; i < renderer->memoryProperties.memoryHeapCount; i += 1)
+	{
+		if (	renderer->memoryProperties.memoryHeaps[i].flags &
+			VK_MEMORY_HEAP_DEVICE_LOCAL_BIT	)
+		{
+			if (renderer->memoryProperties.memoryHeaps[i].size > deviceLocalHeapSize)
```
kg, line 2555: You could probably replace this with SDL_max

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2892,65 +3108,284 @@ static uint8_t VULKAN_INTERNAL_AllocateMemory(
...
+		&allocation->memory
+	);
+
+	if (result != VK_SUCCESS)
+	{
+		/* Uh oh, we couldn't allocate, time to clean up */
+		SDL_free(allocation->freeRegions);
+
+		allocator->allocationCount -= 1;
+		allocator->allocations = SDL_realloc(
```
kg, line 3225: Is this really wise? It could force a new copy of the allocations buffer to be allocated (1 element smaller) and then copied into, which is questionable if we're already low on memory

thatcosmonaut: IMO if the application is so memory-starved that a realloc fails we are already extremely hosed

src/FNA3D_Driver_Vulkan.c
flibitijibibo: With the new defrag work we might actually be able to turn this off entirely - I imagine OS programs using Vulkan will know to fall back to host memory as well.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2892,40 +3292,103 @@ static uint8_t VULKAN_INTERNAL_AllocateMemory(
...
+		(renderer->memoryProperties.memoryTypes[memoryTypeIndex].propertyFlags &
+		VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT) != 0;
+
 	allocator = &renderer->memoryAllocator->subAllocators[memoryTypeIndex];
 	requiredSize = memoryRequirements->memoryRequirements.size;
 
+	if (
+		(buffer == VK_NULL_HANDLE && image == VK_NULL_HANDLE) ||
+		(buffer != VK_NULL_HANDLE && image != VK_NULL_HANDLE)
+	)
```
flibitijibibo, line 3382:
```suggestion
	if (	(buffer == VK_NULL_HANDLE && image == VK_NULL_HANDLE) ||
		(buffer != VK_NULL_HANDLE && image != VK_NULL_HANDLE)	)
```

thatcosmonaut:
After some extensive testing, I have made a few revisions to the defrag strategy:

1) We now walk the descriptor set cache and framebuffer cache when an image view is destroyed to invalidate resources that reference that image. This was causing device loss errors when Vulkan gave us a reused image view handle.

2) Buffer uploads wait on the defrag fence to avoid a data race if the GPU happens to be copying a buffer at the same time as the upload occurs. (In the future we could probably avoid doing this by having buffer uploads also use a staging buffer.)

3) Defrag is now incremental - defrag triggers 5 frames after a resource free and continues every 5 frames until fragmentation is eliminated. This prevents defrag from constantly churning while resources are being loaded/unloaded and smooths out the performance hit.

flibitijibibo (commented):
Only found one style thing! If this survives another sniff test through the catalog this is ready to go.

src/FNA3D_Driver_Vulkan.c
thatcosmonaut: I'm fine with leaving this in as a debug env var if people want to simulate running their games on low-memory devices, but I agree that we can probably just bump this up to 1.0 at the very least now.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2490,6 +2583,28 @@ static uint8_t VULKAN_INTERNAL_DeterminePhysicalDevice(
...
+			{
+				deviceLocalHeapSize = renderer->memoryProperties.memoryHeaps[i].size;
+			}
+		}
+	}
+
+	deviceLocalHeapUsageFactorStr = SDL_GetHint("FNA3D_VULKAN_DEVICE_LOCAL_HEAP_USAGE_FACTOR");
+	if (deviceLocalHeapUsageFactorStr != NULL)
+	{
+		deviceLocalHeapUsageFactor = SDL_max(0, SDL_min(1.0, SDL_atof(deviceLocalHeapUsageFactorStr)));
```
flibitijibibo, line 2602: We may want to have a default for SDL_atof to avoid it accidentally passing 0 when the env var isn't a valid float:
```suggestion
		double factor = SDL_atof(deviceLocalHeapUsageFactorStr);
		if (factor > 0.0 && factor < 1.0)
		{
			deviceLocalHeapUsageFactor = factor;
		}
```

PR #111, opened 2021-05-01
Optimize OpenGL driver in case FNA game doesn't set FNA3D_SETDATAOPTIONS_DISCARD in OPENGL_SetVertexBufferData

Maccraft123:
This is meant to work around it. Probably breaks some games, it can't be default option.
Greatly improves performance on tiling GPUs, almost makes Terraria playable on my Mali G72 machine. Doesn't help immediate mode renderers.

Hack-Suggested-By: Rob Clark


flibitijibibo:
We don't do hacks in our renderer. If anything we probably do the opposite and _avoid_ using discard on drivers where BufferData is expensive and BufferSubData doesn't hit contention problems.

Use Vulkan instead if OpenGL is bottlenecking.

PR #114, opened 2021-07-04
Refactor memory binding to avoid goto

thatcosmonaut:


src/FNA3D_Driver_Vulkan.c
```diff
@@ -2241,12 +2241,6 @@ static uint8_t VULKAN_INTERNAL_FindMemoryType(
 		}
 	}
 
-	FNA3D_LogWarn(
-		"Failed to find memory type %X, required %X, ignored %X",
-		typeFilter,
-		requiredProperties,
-		ignoredProperties
-	);
```
flibitijibibo, line 2249: tbh this warning is the only reason I was able to diagnose this at all - rest looks good but this should really stay.

PR #117, opened 2021-07-28
Found at least one driver based on llvmpipe that did not include anisotropic filtering extension

stuckie:
Implemented checks w.r.t existing S3TC and DXT1 checks.

Specific driver details from FNA3D:
FNA3D Driver: OpenGL
OpenGL Renderer: llvmpipe (LLVM 11.0.0, 256 bits)
OpenGL Driver: 3.1 Mesa 20.2.6
OpenGL Vendor: Mesa/X.org
MojoShader Profile: glsl120

### DIFF
--- a/src/FNA3D_Driver_OpenGL.c
+++ b/src/FNA3D_Driver_OpenGL.c
@@ -175,6 +175,7 @@ typedef struct OpenGLRenderer /* Cast from FNA3D_Renderer* */
 	/* Capabilities */
 	uint8_t supports_s3tc;
 	uint8_t supports_dxt1;
+	uint8_t supports_anisotropic_filtering;
 	int32_t maxMultiSampleCount;
 	int32_t maxMultiSampleCountFormat[21];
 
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_OpenGL.c
```diff
@@ -5536,6 +5550,11 @@ static inline void CheckExtensions(
 	{
 		*supportsDxt1 = 1;
 	}
+
+	if (SDL_strstr(ext, "GL_EXT_texture_filter_anisotropic"))
```
flibitijibibo, line 5554: This needs to check for the ARB extension as well.

src/FNA3D_Driver_OpenGL.c
```diff
@@ -5147,6 +5154,12 @@ static uint8_t OPENGL_SupportsS3TC(FNA3D_Renderer *driverData)
 	return renderer->supports_s3tc;
 }
 
+static uint8_t OPENGL_SupportsAnisotropicFiltering(FNA3D_Renderer *driverData)
```
flibitijibibo, line 5157: This doesn't get exported, so it can be deleted entirely.

flibitijibibo (changes requested):
A couple little things, otherwise good.

stuckie:
Applied changes and here was an additional variant of the driver that hiccupped:

FNA3D Driver: OpenGL
OpenGL Renderer: Gallium 0.4 on llvmpipe (LLVM 3.6, 128 bits)
OpenGL Driver: 3.0 Mesa 11.0.4 (git-)
OpenGL Vendor: VMware, Inc.

PR #118, opened 2021-07-28
Fixed issue with GenColorRenderbuffer and a single sample.

stuckie:
Looks like it should have been the same logic as GenDepthStencilRenderbuffer


flibitijibibo:
Ah yeah, this comes from an assumption on FNA's part where we only use color renderbuffers for multisampling, but it's probably a good idea to check for non-multisampled color buffers too.

Good catch and thanks for the fix!

PR #119, opened 2021-08-06
Initial experimental support for custom SPIRV shaders

GlaireDaggers:
Adds an API for bypassing the Effect framework with custom SPIRV shader modules, using SPIRV-Cross to transpile to target-specific shaders,.


thatcosmonaut:
Just noting here that we have decided to move to a plugin-based system for shader support. Step 1 will be moving the current MojoShader systems into a plugin and then we can start converting this PR to the plugin system.

flibitijibibo:
Closing this since I don't think we'll be taking it in at any point... I do believe someone in the Discord mentioned having a cross compiler that included DXBC support, I would look into that. For upgrading shader models it'll require exposing the shader system, and until PS4 is done I don't think I want to even field ideas since it's a big task (and frankly I'd prefer work time go into either PS4 or the offline cross-compiler first, which will cover 99% of what people actually want, which is just access to other shader languages and not necessarily modern features).

PR #120, opened 2021-08-15
Buffer trace disk writes

thatcosmonaut:
Constantly opening and closing the trace file was murdering performance, so I restructured tracing to write first into a memory buffer and then flush to disk on SwapBuffers and DestroyDevice.

### DIFF
--- a/src/FNA3D_Tracing.c
+++ b/src/FNA3D_Tracing.c
@@ -170,16 +170,17 @@ void FNA3D_Trace_RegisterEffect(FNA3D_Effect *effect, MOJOSHADER_effect *effectD
 }
 #undef TRACE_OBJECT
 
-#define WRITE(val) SDL_memcpy((uint8_t*)traceBuffer + traceBufferCurrentSize, &val, sizeof(val)); traceBufferCurrentSize += sizeof(val);
-#define WRITEMEM(ptr, len) SDL_memcpy((uint8_t*)traceBuffer + traceBufferCurrentSize, ptr, len); traceBufferCurrentSize += len;
+#define CHECK_AND_FLUSH_BUFFER(len) if (traceBufferCurrentSize + len > traceBufferSize) { FNA3D_Trace_FlushMemory(); }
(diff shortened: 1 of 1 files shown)


src/FNA3D_Tracing.c
```diff
@@ -170,26 +170,41 @@ void FNA3D_Trace_RegisterEffect(FNA3D_Effect *effect, MOJOSHADER_effect *effectD
 }
 #undef TRACE_OBJECT
 
-#define WRITE(val) ops->write(ops, &val, sizeof(val), 1)
+#define CHECK_AND_FLUSH_BUFFER(len) if (traceBufferCurrentSize + len > traceBufferSize) { FNA3D_Trace_FlushMemory(); }
+#define WRITE(val) CHECK_AND_FLUSH_BUFFER(sizeof(val)); SDL_memcpy((uint8_t*)traceBuffer + traceBufferCurrentSize, &val, sizeof(val)); traceBufferCurrentSize += sizeof(val);
+#define WRITEMEM(ptr, len) CHECK_AND_FLUSH_BUFFER(len); SDL_memcpy((uint8_t*)traceBuffer + traceBufferCurrentSize, ptr, len); traceBufferCurrentSize += len;
```
flibitijibibo, line 175: We can probably expand these now...

```suggestion
#define CHECK_AND_FLUSH_BUFFER(len) \
	if (traceBufferCurrentSize + len > traceBufferSize) \
	{ \
		FNA3D_Trace_FlushMemory(); \
	}
#define WRITE(val) \
	CHECK_AND_FLUSH_BUFFER(sizeof(val)) \
	SDL_memcpy((uint8_t*) traceBuffer + traceBufferCurrentSize, &val, sizeof(val)); \
	traceBufferCurrentSize += sizeof(val);
#define WRITEMEM(ptr, len) \
	CHECK_AND_FLUSH_BUFFER(len) \
	SDL_memcpy((uint8_t*) traceBuffer + traceBufferCurrentSize, ptr, len); \
	traceBufferCurrentSize += len;
```

flibitijibibo (changes requested):
Just the one thing, otherwise fine

PR #121, opened 2021-08-17
Add support for BC7 textures

kg:
Putting this here so other people can test it in the hopes that we can eventually land it. FNA changes also necessary to consume this but they're pretty obvious. Will clean up whitespace and clean up the vcxproj damage before flipping this out of draft status (i love vc++ demanding a retarget before you can do anything). Also need to figure out how to handle feature detection for this and surface an error if bc7 isn't available on the target, not sure if I can manually test that locally. Also still need to figure out how this works in OpenGL (probably just an extension).

Instead of the chained == in the various ifs should we introduce a convenience 'is_block_compressed_format(x)' function to use in all those if statements?

Should we roll in SRGB surface formats along with this since adding those touches a lot of the same code and has some of the same problems (feature detection, etc)?


flibitijibibo:
This should be rebased against the latest tip (we should probably add the GL support as well).

flibitijibibo:
This should be rebased on master to account for the new SRGB entries.

flibitijibibo:
Bunch of stuff got merged over the weekend. D3D11 multiwindow will probably shake even more stuff up so this should be rebased ASAP.

kg:
Still need to do some more testing, but I rebased and added the relevant OpenGL enums.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -1625,9 +1625,11 @@ static VkFormat XNAToVK_SurfaceFormat[] =
 	VK_FORMAT_R16G16_SFLOAT,		/* SurfaceFormat.HalfVector2 */
 	VK_FORMAT_R16G16B16A16_SFLOAT,		/* SurfaceFormat.HalfVector4 */
 	VK_FORMAT_R16G16B16A16_SFLOAT,		/* SurfaceFormat.HdrBlendable */
-	VK_FORMAT_B8G8R8A8_UNORM,		/* SurfaceFormat.ColorBgraEXT */
+	VK_FORMAT_R8G8B8A8_UNORM,		/* SurfaceFormat.ColorBgraEXT */
```
thatcosmonaut, line 1628: Just curious, is this getting swizzled or something?

flibitijibibo: It shouldn't be, this is based on input data - this line should be undone.

src/FNA3D_Driver_D3D11.c
```diff
@@ -310,7 +311,8 @@ static DXGI_FORMAT XNAToD3D_TextureFormat[] =
 	DXGI_FORMAT_B8G8R8A8_UNORM,	/* SurfaceFormat.ColorBgraEXT */
 	DXGI_FORMAT_R8G8B8A8_UNORM_SRGB,/* SurfaceFormat.ColorSrgbEXT */
 	DXGI_FORMAT_BC3_UNORM_SRGB,	/* SurfaceFormat.Dxt5SrgbEXT */
-};
+	DXGI_FORMAT_BC7_UNORM, /* SurfaceFormat.BC7EXT */
+	DXGI_FORMAT_BC7_UNORM_SRGB,	/* SurfaceFormat.BC7SrgbEXT */};
```
flibitijibibo, line 315:
```suggestion
	DXGI_FORMAT_BC7_UNORM_SRGB,	/* SurfaceFormat.BC7SrgbEXT */
};
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -367,6 +368,8 @@ static int32_t XNAToGL_TextureInternalFormat[] =
 	GL_RGBA8,				/* SurfaceFormat.ColorBgraEXT */
 	GL_SRGB_ALPHA_EXT,			/* SurfaceFormat.ColorSrgbEXT */
 	GL_COMPRESSED_SRGB_ALPHA_S3TC_DXT5_EXT,	/* SurfaceFormat.Dxt5SrgbEXT */
+	GL_COMPRESSED_RGBA_BPTC_UNORM_EXT,	/* SurfaceFormat.BC7EXT */
+	GL_COMPRESSED_SRGB_ALPHA_BPTC_UNORM_EXT,/* SurfaceFormat.BC7SrgbEXT */
```
flibitijibibo, line 372: Be sure to fill in the other enums - I think they'll just be NOPE entries.

flibitijibibo (changes requested):
Also need the project file changes undone but this should be really straightforward.

kg:
Verified working in D3D11, Vulkan and OpenGL as of the latest commit (at least for my test case).

flibitijibibo (approved):
Latest looks good. Will merge probably at the same time as SDL 2.0.20, gives some more time to test.

PR #123, opened 2021-08-18
D3D11: Try to discard target textures before setting them as targets

### LINKED ISSUE #122: D3D11 SetRenderTargets needs an inverse of RestoreTargetTextures
An example warning:

```
D3D11 WARNING: ID3D11DeviceContext::OMSetRenderTargets: Resource being set to OM RenderTarget slot 0 is still bound on input! [ STATE_SETTING WARNING #9: DEVICE_OMSETRENDERTARGETS_HAZARD]
D3D11 WARNING: ID3D11DeviceContext::OMSetRenderTargets[AndUnorderedAccessViews]: Forcing PS shader resource slot 1 to NULL. [ STATE_SETTING WARNING #7: DEVICE_PSSETSHADERRESOURCES_HAZARD]
```

Basically what happens is, Textures[1] might be set to a RenderTarget, then an app might do:

```
GraphicsDevice.SetRenderTarget(thatTexture);

// Then later, but before a draw...
GraphicsDevice.Textures[1] = null; /* AKA: Not thatTexture anymore */
```

Even though the state will be valid by the time it Draws, you'll still get a warning from D3D. We just need to do the opposite of what our RestoreTargetTextures does here.

flibitijibibo:
Hopefully fixes #122

### DIFF
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -2141,6 +2141,82 @@ static void D3D11_ApplyVertexBufferBindings(
 
 /* Render Targets */
 
+static void D3D11_INTERNAL_DiscardTargetTextures(
+	D3D11Renderer *renderer,
+	ID3D11RenderTargetView **views,
+	int32_t numViews
+) {
+	/* For textures that are still bound while this target is about to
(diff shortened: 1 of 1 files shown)


TheSpydog (approved):
LGTM!

PR #127, opened 2021-09-11
Support srgb render targets

bartwe:


flibitijibibo (changes requested):
See review for #85

include/FNA3D.h
```diff
@@ -1478,6 +1479,9 @@ FNA3DAPI uint8_t FNA3D_SupportsHardwareInstancing(FNA3D_Device *device);
 /* Returns 1 if the renderer natively supports asynchronous buffer writing. */
 FNA3DAPI uint8_t FNA3D_SupportsNoOverwrite(FNA3D_Device *device);
 
+/* Returns 1 if the renderer natively supports a SRGB frame buffer. */
+FNA3DAPI uint8_t FNA3D_SupportsSRGBFrameBuffer(FNA3D_Device *device);
```
flibitijibibo, line 1483: Mostly following the XNA terminology rather than GL...

```suggestion
/* Returns 1 if the renderer natively supports SRGB render targets. */
FNA3DAPI uint8_t FNA3D_SupportsSRGBRenderTargets(FNA3D_Device *device);
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -106,6 +106,7 @@ struct OpenGLRenderbuffer /* Cast from FNA3D_Renderbuffer* */
 {
 	GLuint handle;
 	OpenGLRenderbuffer *next; /* linked list */
+	FNA3D_SurfaceFormat format;
```
flibitijibibo, line 109: Link needs to be at the bottom

```suggestion
	FNA3D_SurfaceFormat format;
	OpenGLRenderbuffer *next; /* linked list */
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -2546,13 +2558,17 @@ static void OPENGL_SetRenderTargets(
 		renderer->renderTargetBound = 1;
 	}
 
+	uint8_t isSrgb = 0;
+
+
```
flibitijibibo, line 2563:
```suggestion
	uint8_t isSrgb = 0;

```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -5548,6 +5582,9 @@ static inline void CheckExtensions(
 		SDL_strstr(ext, "GL_EXT_texture_filter_anisotropic") ||
 		SDL_strstr(ext, "GL_ARB_texture_filter_anisotropic")
 	);
+	uint8_t srgbFrameBuffer = (
+		SDL_strstr(ext, "GL_EXT_framebuffer_sRGB")
```
flibitijibibo, line 5586: Did you happen to run into a hardware config where this didn't work? I'm surprised this got tripped, but at the same time we do support really old framebuffer extensions so this is probably a good idea...

src/FNA3D_Driver_OpenGL.c
```diff
@@ -5536,7 +5569,8 @@ static inline void CheckExtensions(
 	const char *ext,
 	uint8_t *supportsS3tc,
 	uint8_t *supportsDxt1,
-	uint8_t *supportsAnisotropicFiltering
+	uint8_t *supportsAnisotropicFiltering,
+	uint8_t *SupportsSRGBFrameBuffer
```
flibitijibibo, line 5573:
```suggestion
	uint8_t *supportsSRGBFrameBuffer
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -11498,6 +11505,11 @@ static uint8_t VULKAN_SupportsNoOverwrite(FNA3D_Renderer *driverData)
 	return 1;
 }
 
+static uint8_t VULKAN_SupportsSRGBFrameBuffer(FNA3D_Renderer *driverData)
+{
+	return 1;
```
flibitijibibo, line 11510: We should probably check this with vkGetPhysicalDeviceFormatProperties (see S3TC's query, but with different bitflags).

src/FNA3D_Driver_D3D11.c
```diff
@@ -4582,6 +4586,11 @@ static uint8_t D3D11_SupportsNoOverwrite(FNA3D_Renderer *driverData)
 	return 1;
 }
 
+static uint8_t D3D11_SupportsSRGBFrameBuffer(FNA3D_Renderer *driverData)
+{
+	return 1;
```
flibitijibibo, line 4591: Should probably do ID3D11Device_CheckFormatSupport, similar to S3TC (which I just noticed uses the return value incorrectly...)

flibitijibibo (changes requested):
A handful of architecture things, and details for a fully-conformant implementation.

src/FNA3D_Driver_OpenGL.c
bartwe: No i was merely following the pattern of the other feature detections

bartwe:
Ah, well accepted those edits to have them flow back into my git so i could work on them and was planning to pick up those two open points later tonight

flibitijibibo:
Sounds good - Kate should be able to review and test, and having both users of this feature actively looking at the latest revision will be important.

PR #131, opened 2021-11-13
Vulkan: Fix SetBufferData overwriting in-flight None and Discard submissions

thatcosmonaut:
[`03d90b4`](https://github.com/FNA-XNA/FNA3D/commit/f54ebc2e68f509209d48e5eca374d14f6f707307) fixed an issue where the sub-buffer index was never reset, causing unbounded sub-buffer growth. Unfortunately, this also meant that `SetBufferData` was now actually capable of overwriting already-used sub-buffers. As a consequence, it was selecting sub-buffers that were still being used in rendering and overwriting their data.

The Vulkan renderer is unique in that we begin processing frames while the previously submitted frame is still rendering. This means that we need to be extra careful in how we select `SetBufferData` sub-buffers.

The process is now as follows:

If NOOVERWRITE is set, we assume the user knows what they're doing and copy using the current sub buffer index even if the buffer might be in-flight. This is actually XNA behavior!

If NONE or DISCARD is set, we check if the buffer was bound either this frame or the previous frame. If so, we increment the index until we find a sub-buffer that is unbound. Otherwise we use the current sub-buffer index.

If no valid sub-buffer was found, we allocate a new one.

If NONE is set and the buffer was bound this frame, we copy its data from the old sub-buffer to the next one.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -6818,7 +6818,8 @@ static void VULKAN_INTERNAL_SetBufferData(
 	 * If so, we start at sub-buffer 0 and increment the index until we find a sub-buffer that is unbound.
 	 * Otherwise we use the current sub-buffer index.
 	 */
-	if ((options != FNA3D_SETDATAOPTIONS_NOOVERWRITE && (vulkanBuffer->bound || vulkanBuffer->boundSubmitted)))
+	if (	options != FNA3D_SETDATAOPTIONS_NOOVERWRITE &&
+		(vulkanBuffer->bound || vulkanBuffer->boundSubmitted)	)
 	{
 		CURIDX = 0;
 		while (CURIDX < vulkanBuffer->subBufferCount && SUBBUF->bound != -1)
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_Vulkan.c
```diff
@@ -6808,45 +6808,26 @@ static void VULKAN_INTERNAL_SetBufferData(
...
-
-	/*
-	 * If buffer was bound and options is NONE or DISCARD
-	 * find the next available unbound sub-buffer
+	/* If NONE or DISCARD is set, we check if the buffer was bound either this frame or the previous frame.
+	 * If so, we start at sub-buffer 0 and increment the index until we find a sub-buffer that is unbound.
+	 * Otherwise we use the current sub-buffer index.
 	 */
-	if (vulkanBuffer->bound)
+	if ((options != FNA3D_SETDATAOPTIONS_NOOVERWRITE && (vulkanBuffer->bound || vulkanBuffer->boundSubmitted)))
```
flibitijibibo, line 6821: Had a bit of trouble parsing this one:

```suggestion
	if (	options != FNA3D_SETDATAOPTIONS_NOOVERWRITE &&
		(vulkanBuffer->bound || vulkanBuffer->boundSubmitted)	)
```

flibitijibibo (commented):
The changes make sense to me, but we'll probably want to redo the commit message here... Once that's done this can be squashed in.

PR #132, opened 2021-11-16
Added ETC2 texture format support for iOS

ryancheung:


flibitijibibo:
As far as patches go this is very well-written! A new `FNA3D_SupportsETC2` would be needed to query support but other than that this makes sense to me.

I'll defer to someone with more mobile experience for the actual content and functionality, since in my head I always thought ASTC was supposed to be the mobile standard going forward... but I have limited experience here.

We have a number of other texture extensions pending, so what I'll most likely do because this adds more than the other is try to merge them in the following order:

1. SRGB
2. BC7
3. ETC2 (or whatever seems best if the mobile experts prefer something else)

include/FNA3D.h
```diff
@@ -132,7 +132,10 @@ typedef enum FNA3D_SurfaceFormat
 	FNA3D_SURFACEFORMAT_HALFVECTOR2,
 	FNA3D_SURFACEFORMAT_HALFVECTOR4,
 	FNA3D_SURFACEFORMAT_HDRBLENDABLE,
-	FNA3D_SURFACEFORMAT_COLORBGRA_EXT
+	FNA3D_SURFACEFORMAT_COLORBGRA_EXT,
+	FNA3D_SURFACEFORMAT_RGB8ETC2,
```
kg, line 136: my preference would be for ETC2 to be at the front of the format name, not the end

src/FNA3D_Driver.h
```diff
@@ -202,7 +207,9 @@ static inline int32_t BytesPerRow(
 
 	if (	format == FNA3D_SURFACEFORMAT_DXT1 ||
 		format == FNA3D_SURFACEFORMAT_DXT3 ||
-		format == FNA3D_SURFACEFORMAT_DXT5	)
+		format == FNA3D_SURFACEFORMAT_DXT5 ||
+		format == FNA3D_SURFACEFORMAT_RGB8ETC2 ||
+		format == FNA3D_SURFACEFORMAT_RGB8A1ETC2	)
```
kg, line 212: is FNA3D_SURFACEFORMAT_RGBA8ETC2 meant to be missing here?

ryancheung:
Yeah. ASTC should be the mobile standard going forward. ASTC support should be added too.

According to the doc in [Unity](https://docs.unity3d.com/Manual/class-TextureImporterOverride.html). ETC2 should have better device compability than ASTC. Apple devices support ETC/ETC2 formats starting with A7 chip (2013). For ASTC, A8 chip (2014) or above is required. Android devices support ETC2 with GLES 3.0+ GPU and ASTC with OpenGL 3.1+(or parts of 3.0) GPU.

flibitijibibo:
At least for our purposes we only support A8 and above anyway, so that'd be no problem. Android by itself I'd be hesitant to add anything for because in general we stay as far away from it as possible.

If we _had_ to factor it in I would base it on the Vulkan landscape rather than GLES - we could hypothetically check a database for OES_texture_compression_astc but odds are we'd get more realistic data from checking gpuinfo.org's Vulkan Android database for ASTC support. My guess is devices made within the last ~5 years will have it, and for Android that's a _huge_ data set.

flibitijibibo:
Not sure why I've not looked into this before... just checked Vulkan Android and ASTC is almost universally supported:

https://vulkan.gpuinfo.org/listoptimaltilingformats.php?platform=android

My guess is the remaining % is non-mobile.

ryancheung:
Like Godot 4.0 do with Vulkan only in the beginning.
So that means it is possbile now for us to add Android support with Vulkan in FNA?

flibitijibibo:
Android support is off the table - always has been, always will be. But it's likely the support matrix for Android Vulkan will be similar to what comes after it, i.e. Fuchsia or alternatives like Plasma Mobile and Phosh.

PR #134, opened 2021-12-04
Remove Metal renderer

TheSpydog:


flibitijibibo (approved):
Looks good, all the other projects don't touch these files/defines so you should be set!

PR #135, opened 2021-12-22
Vulkan: Multi-window support

thatcosmonaut:
This patch uses a dictionary of window handles to swapchains to blit the faux-backbuffer to the appropriate swapchain to handle multiple windows on the same device.

Some outstanding concerns: Now that we aren't using a one swapchain assumption, the initialization flow feels slightly awkward. This is related to the special cases for `deviceWindowHandle` in the swapchain functions. We might want to consider initializing with a dummy surface similar to the `PrepareWindowAttributes` flow and then handling windows/surfaces/swapchains as always-bundled units, but I'm not sure if this would also be awkward.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -6145,7 +6145,9 @@ static void VULKAN_INTERNAL_SubmitCommands(
 	VkSemaphore semaphores[2];
 	VkSubmitInfo submitInfo;
 	uint32_t i, j;
-	VkResult result, acquireResult, presentResult = VK_SUCCESS;
+	VkResult result;
+	VkResult acquireResult = VK_SUCCESS;
+	VkResult presentResult = VK_SUCCESS;
 	uint8_t acquireSuccess = 0;
 	uint8_t performDefrag = 0;
 	uint8_t createSwapchainResult = 0;
@@ -6453,7 +6455,7 @@ static void VULKAN_INTERNAL_SubmitCommands(
 	/* Now that commands are totally done, check if we need new swapchain */
 	if (present)
 	{
-		if (	!validSwapchainExists ||	
+		if (	!validSwapchainExists ||
 			acquireResult == VK_ERROR_OUT_OF_DATE_KHR ||
 			acquireResult == VK_SUBOPTIMAL_KHR ||
 			presentResult == VK_ERROR_OUT_OF_DATE_KHR ||
@@ -6566,6 +6568,7 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 
 	if (swapchainData != NULL)
 	{
+		SDL_free(swapchainData);
 		FNA3D_LogError("Swapchain already exists for this window! Destroy it first!");
 		return CREATE_SWAPCHAIN_FAIL;
 	}
@@ -6579,6 +6582,7 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 		renderer->instance,
 		&swapchainData->surface
 	)) {
+		SDL_free(swapchainData);
 		FNA3D_LogError(
 			"SDL_Vulkan_CreateSurface failed: %s",
 			SDL_GetError()
@@ -6595,6 +6599,20 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 
 	if (swapchainSupport == 0)
 	{
+		renderer->vkDestroySurfaceKHR(
+			renderer->instance,
+			swapchainData->surface,
+			NULL
+		);
+		if (swapchainSupportDetails.formatsLength > 0)
+		{
+			SDL_free(swapchainSupportDetails.formats);
+		}
+		if (swapchainSupportDetails.presentModesLength > 0)
+		{
+			SDL_free(swapchainSupportDetails.presentModes);
+		}
+		SDL_free(swapchainData);
 		FNA3D_LogError("Surface does not support swapchain creation!");
 		return CREATE_SWAPCHAIN_FAIL;
 	}
@@ -6628,6 +6646,11 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 			swapchainSupportDetails.formatsLength,
 			&swapchainData->surfaceFormat
 		)) {
+			renderer->vkDestroySurfaceKHR(
+				renderer->instance,
+				swapchainData->surface,
+				NULL
+			);
 			if (swapchainSupportDetails.formatsLength > 0)
 			{
 				SDL_free(swapchainSupportDetails.formats);
@@ -6636,6 +6659,7 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 			{
 				SDL_free(swapchainSupportDetails.presentModes);
 			}
+			SDL_free(swapchainData);
 			FNA3D_LogError("Device does not support swap chain format");
 			return CREATE_SWAPCHAIN_FAIL;
 		}
@@ -6647,6 +6671,11 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 		swapchainSupportDetails.presentModesLength,
 		&swapchainData->presentMode
 	)) {
+		renderer->vkDestroySurfaceKHR(
+			renderer->instance,
+			swapchainData->surface,
+			NULL
+		);
 		if (swapchainSupportDetails.formatsLength > 0)
 		{
 			SDL_free(swapchainSupportDetails.formats);
@@ -6655,6 +6684,7 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 		{
 			SDL_free(swapchainSupportDetails.presentModes);
 		}
+		SDL_free(swapchainData);
 		FNA3D_LogError("Device does not support swap chain present mode");
 		return CREATE_SWAPCHAIN_FAIL;
 	}
@@ -6675,6 +6705,11 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 		if (	swapchainSupportDetails.capabilities.currentExtent.width == 0 ||
 			swapchainSupportDetails.capabilities.currentExtent.height == 0)
 		{
+			renderer->vkDestroySurfaceKHR(
+				renderer->instance,
+				swapchainData->surface,
+				NULL
+			);
 			if (swapchainSupportDetails.formatsLength > 0)
 			{
 				SDL_free(swapchainSupportDetails.formats);
@@ -6683,6 +6718,7 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 			{
 				SDL_free(swapchainSupportDetails.presentModes);
 			}
+			SDL_free(swapchainData);
 			return CREATE_SWAPCHAIN_SURFACE_ZERO;
 		}
 
@@ -6702,7 +6738,11 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 		}
 		else
 		{
-			FNA3D_LogError("No fallback swapchain size available!");
+			renderer->vkDestroySurfaceKHR(
+				renderer->instance,
+				swapchainData->surface,
+				NULL
+			);
 			if (swapchainSupportDetails.formatsLength > 0)
 			{
 				SDL_free(swapchainSupportDetails.formats);
@@ -6711,6 +6751,8 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 			{
 				SDL_free(swapchainSupportDetails.presentModes);
 			}
+			SDL_free(swapchainData);
+			FNA3D_LogError("No fallback swapchain size available!");
 			return CREATE_SWAPCHAIN_FAIL;
 		}
 	}
@@ -6778,6 +6820,11 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 	if (!swapchainData->images)
 	{
 		SDL_OutOfMemory();
+		renderer->vkDestroySurfaceKHR(
+			renderer->instance,
+			swapchainData->surface,
+			NULL
+		);
 		if (swapchainSupportDetails.formatsLength > 0)
 		{
 			SDL_free(swapchainSupportDetails.formats);
@@ -6786,6 +6833,7 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 		{
 			SDL_free(swapchainSupportDetails.presentModes);
 		}
+		SDL_free(swapchainData);
 		return CREATE_SWAPCHAIN_FAIL;
 	}
 
@@ -6795,6 +6843,11 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 	if (!swapchainData->views)
 	{
 		SDL_OutOfMemory();
+		renderer->vkDestroySurfaceKHR(
+			renderer->instance,
+			swapchainData->surface,
+			NULL
+		);
 		if (swapchainSupportDetails.formatsLength > 0)
 		{
 			SDL_free(swapchainSupportDetails.formats);
@@ -6803,6 +6856,7 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 		{
 			SDL_free(swapchainSupportDetails.presentModes);
 		}
+		SDL_free(swapchainData);
 		return CREATE_SWAPCHAIN_FAIL;
 	}
 
@@ -6812,6 +6866,11 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 	if (!swapchainData->resourceAccessTypes)
 	{
 		SDL_OutOfMemory();
+		renderer->vkDestroySurfaceKHR(
+			renderer->instance,
+			swapchainData->surface,
+			NULL
+		);
 		if (swapchainSupportDetails.formatsLength > 0)
 		{
 			SDL_free(swapchainSupportDetails.formats);
@@ -6820,6 +6879,7 @@ static CreateSwapchainResult VULKAN_INTERNAL_CreateSwapchain(
 		{
 			SDL_free(swapchainSupportDetails.presentModes);
 		}
+		SDL_free(swapchainData);
 		return CREATE_SWAPCHAIN_FAIL;
 	}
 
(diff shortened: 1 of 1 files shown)


thatcosmonaut:
Tweaked the initialization flow - now every window, including the default window, has its own surface and swapchain via the window hash lookup.

flibitijibibo:
The hash table seems excessive for something that won't have that much data... would it make sense to just have an array of swapchain info and assign them to windows via SDL_SetWindowData?

thatcosmonaut:
It's a 1-to-1 association so that would probably work... let me think about this.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -5983,10 +6080,11 @@ static void VULKAN_INTERNAL_ResetDescriptorSetData(VulkanRenderer *renderer)
...
 static void VULKAN_INTERNAL_SwapChainBlit(
 	VulkanRenderer *renderer,
-	FNA3D_Rect *sourceRectangle,
+	VulkanSwapchainData *swapchainData,
+	FNA3D_Rect * sourceRectangle,
 	FNA3D_Rect *destinationRectangle,
-	uint32_t swapChainImageIndex
-) {
+	uint32_t swapchainImageIndex)
+{
```
flibitijibibo, line 6087:
```suggestion
	uint32_t swapchainImageIndex
) {
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6150,23 +6247,45 @@ static void VULKAN_INTERNAL_SubmitCommands(
 			mode.refresh_rate = 60;
 		}
 
+		swapchainData = WindowSwapchainHashArray_Fetch(
+			&renderer->windowSwapchainTable,
+			windowHandle
+		);
+
+		/* TODO: create the swapchain here if NULL */
```
flibitijibibo, line 6255: Looks like this TODO is obsolete now?

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6490,256 +6611,190 @@ static void VULKAN_INTERNAL_FlushCommandsAndPresent(
...
-	swapChainCreateInfo.preTransform = swapChainSupportDetails.capabilities.currentTransform;
-	swapChainCreateInfo.compositeAlpha = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
-	swapChainCreateInfo.presentMode = presentMode;
-	swapChainCreateInfo.clipped = VK_TRUE;
-	swapChainCreateInfo.oldSwapchain = VK_NULL_HANDLE;
+		swapchainData->imageCount = SDL_max(swapchainData->imageCount, 3);
+	}
+
+	/* Each swapchain must have its own surface.
+	 */
```
flibitijibibo, line 6659:
```suggestion
	/* Each swapchain must have its own surface. */
```

flibitijibibo (changes requested):
Minor details aside from the hashmap work.

thatcosmonaut:
Alright, it uses `SDL_SetWindowData` and `SDL_GetWindowData` instead of a hash now.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6150,23 +6158,37 @@ static void VULKAN_INTERNAL_SubmitCommands(
 			mode.refresh_rate = 60;
 		}
 
+		swapchainData = (VulkanSwapchainData*)SDL_GetWindowData(windowHandle, WINDOW_SWAPCHAIN_DATA);
```
flibitijibibo, line 6161:
```suggestion
		swapchainData = (VulkanSwapchainData*) SDL_GetWindowData(windowHandle, WINDOW_SWAPCHAIN_DATA);
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6150,23 +6158,37 @@ static void VULKAN_INTERNAL_SubmitCommands(
...
+
+		if (swapchainData == NULL)
+		{
+			if (VULKAN_INTERNAL_CreateSwapchain(renderer, windowHandle) != CREATE_SWAPCHAIN_SUCCESS)
+			{
+				FNA3D_LogError("Failed to create swapchain for window handle: %p", windowHandle);
+				return;
+			}
+
+			swapchainData = (VulkanSwapchainData*)SDL_GetWindowData(windowHandle, WINDOW_SWAPCHAIN_DATA);
```
flibitijibibo, line 6171:
```suggestion
			swapchainData = (VulkanSwapchainData*) SDL_GetWindowData(windowHandle, WINDOW_SWAPCHAIN_DATA);
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6490,256 +6514,185 @@ static void VULKAN_INTERNAL_FlushCommandsAndPresent(
...
-	if (!VULKAN_INTERNAL_QuerySwapChainSupport(
-		renderer,
-		renderer->physicalDevice,
-		renderer->surface,
-		&swapChainSupportDetails
-	)) {
-		FNA3D_LogError("Device does not support swap chain creation");
-		return CREATE_SWAPCHAIN_FAIL;
-	}
+	swapchainData = (VulkanSwapchainData*)SDL_GetWindowData(windowHandle, WINDOW_SWAPCHAIN_DATA);
```
flibitijibibo, line 6527:
```suggestion
	swapchainData = (VulkanSwapchainData*) SDL_GetWindowData(windowHandle, WINDOW_SWAPCHAIN_DATA);
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6490,256 +6514,185 @@ static void VULKAN_INTERNAL_FlushCommandsAndPresent(
...
-static void VULKAN_INTERNAL_DestroySwapchain(VulkanRenderer *renderer)
-{
+static void VULKAN_INTERNAL_DestroySwapchain(
+	VulkanRenderer *renderer,
+	void *windowHandle
+) {
 	uint32_t i;
+	VulkanSwapchainData *swapchainData;
+
+	swapchainData = (VulkanSwapchainData*)SDL_GetWindowData(windowHandle, WINDOW_SWAPCHAIN_DATA);
```
flibitijibibo, line 6689:
```suggestion
	swapchainData = (VulkanSwapchainData*) SDL_GetWindowData(windowHandle, WINDOW_SWAPCHAIN_DATA);
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -161,6 +161,8 @@ static inline void CreateDeviceExtensionArray(
 
 #define DEFAULT_PIPELINE_CACHE_FILE_NAME "FNA3D_Vulkan_PipelineCache.blob"
 
+#define WINDOW_SWAPCHAIN_DATA "swapchainData"
```
flibitijibibo, line 164: Might need something more obvious for anyone inspecting via debugger:

```suggestion
#define WINDOW_SWAPCHAIN_DATA "FNA3D_VulkanSwapchain"
```

This will help differentiate it from, say, a D3D11 swapchain pair whenever that gets supported.

flibitijibibo (approved):
Few more tiny things, otherwise this looks fine. We'll want to get this in the hands of Guard Crush and YoYo for live testing. If we think D3D11 will be as straightforward as this we can hold off until both are ready to look at.

flibitijibibo (approved):
Should be okay. Aside from the aforementioned: @kg, I believe this affects the SRGB extension so this may be worth checking even with a single window.

thatcosmonaut:
Noticed a validation error when testing Flotilla, might want to hold off on merging until I take care of that.

thatcosmonaut:
Validation error fixed, I wasn't querying the surface for swapchain support.

thatcosmonaut:
Murder Miners trace lit up some backbuffer resize issues, fixing those.

thatcosmonaut:
We've got a problem - I've resolved some swapchain management issues but it seems like `ResetBackbuffer` is triggering before the window is done resizing, so the surface query parameters are outdated and we try to generate a swapchain that is larger than the window. This is on Fedora 35 Wayland. Happening on master as well, if you run the Murder Miners trace you'll see a Warn about swapchain sizing and the game will only draw at the original size before window resize.

flibitijibibo:
Probably related:

https://github.com/libsdl-org/SDL/issues/4595

https://github.com/libsdl-org/SDL/issues/4563

thatcosmonaut:
Alright, looks like this isn't our renderer's fault at least. I'll re-add the extents check warn.

thatcosmonaut:
Confirmed that SDL_VIDEODRIVER=wayland does not exhibit this behavior, so this is an XWayland issue for sure.

thatcosmonaut:
Ugh it was my fault after all, had a bad realloc in there. Not really sure why the swapchain size is wrong on master but it's fixed on both XWayland and Wayland with this most recent commit.

thatcosmonaut:
Gonna run some traces again. This patchset has ended up being a pretty large improvement to Vulkan swapchain management in general, I'm thinking we should maybe merge this in after testing and I can rebase my buffer sync rewrite on this once it's in.

thatcosmonaut:
Traces look fine, will wait for another review.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6490,256 +6526,345 @@ static void VULKAN_INTERNAL_FlushCommandsAndPresent(
...
+		{
+			SDL_free(swapchainSupportDetails.presentModes);
+		}
 		return CREATE_SWAPCHAIN_FAIL;
 	}
 
-	extent = ChooseSwapExtent(
-		renderer->deviceWindowHandle,
-		swapChainSupportDetails.capabilities
+	/* TODO: validate swapchain extents */
```
flibitijibibo, line 6639: Remaining TODO. In particular, the 0,0 extent is actually really important for surface suspension on Windows and Wayland, so we'll need to reincorporate this where it's been removed.

Other than that this is all fine.

thatcosmonaut:
Re-added zero size surface handling and tightened up the swapchain handling in SwapBuffers a bit more for failure cases.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -38,6 +38,8 @@
 #include <SDL.h>
 #include <SDL_vulkan.h>
 
+#define clamp(val, min, max) SDL_max(min, SDL_min(val, max))
```
flibitijibibo, line 41: Might need a less generic name, though I doubt it will collide with anything...

src/FNA3D_Driver_Vulkan.c
```diff
@@ -229,7 +233,7 @@ typedef enum CreateSwapchainResult
 {
 	CREATE_SWAPCHAIN_FAIL,
 	CREATE_SWAPCHAIN_SUCCESS,
-	CREATE_SWAPCHAIN_SURFACE_ZERO,
+	CREATE_SWAPCHAIN_SURFACE_ZERO
```
flibitijibibo, line 236: Can probably bring back the comma if only to shrink the diff size

```suggestion
	CREATE_SWAPCHAIN_SURFACE_ZERO,
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -1007,6 +1011,22 @@ static inline void PipelineLayoutHashArray_Insert(
...
+	VkSurfaceFormatKHR surfaceFormat;
+	VkPresentModeKHR presentMode;
+	VkFormat swapchainFormat;
+	VkComponentMapping swapchainSwizzle;
+	VkExtent2D extent;
+	VkImage *images;
+	VulkanResourceAccessType *resourceAccessTypes;
+	VkImageView *views;
+	uint32_t imageCount;
+	void *windowHandle;
```
flibitijibibo, line 1027: Cleaning this up as best as I can...
```suggestion
	/* Window surface */
	VkSurfaceKHR surface;
	VkSurfaceFormatKHR surfaceFormat;
	void *windowHandle;

	/* Swapchain for window surface */
	VkSwapchainKHR swapchain;
	VkFormat swapchainFormat;
	VkComponentMapping swapchainSwizzle;
	VkPresentModeKHR presentMode;

	/* Swapchain images */
	VkExtent2D extent;
	VkImage *images;
	VkImageView *views;
	VulkanResourceAccessType *resourceAccessTypes;
	uint32_t imageCount;
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2060,6 +2069,20 @@ static uint8_t VULKAN_INTERNAL_QuerySwapChainSupport(
...
+	renderer->vkGetPhysicalDeviceSurfaceSupportKHR(
+		physicalDevice,
+		renderer->queueFamilyIndex,
+		surface,
+		&supportsPresent
+	);
+
+	if (!supportsPresent)
+	{
+		FNA3D_LogError("This surface does not support presenting!");
```
flibitijibibo, line 2083: This can warn, will error later
```suggestion
		FNA3D_LogWarn("This surface does not support presenting!");
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2060,6 +2069,20 @@ static uint8_t VULKAN_INTERNAL_QuerySwapChainSupport(
 	SwapChainSupportDetails *outputDetails
 ) {
 	VkResult result;
+	VkBool32 supportsPresent;
+
+	renderer->vkGetPhysicalDeviceSurfaceSupportKHR(
```
flibitijibibo, line 2074: I think we check this somewhere else... does one make the other redundant by chance?

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2269,7 +2292,7 @@ static uint8_t VULKAN_INTERNAL_IsDeviceSuitable(
 	uint8_t *deviceRank
 ) {
 	uint32_t queueFamilyCount, queueFamilyRank, queueFamilyBest;
-	SwapChainSupportDetails swapChainSupportDetails;
+	SwapChainSupportDetails swapchainSupportDetails;
```
flibitijibibo, line 2295: Nit: Totally agree with the new capitalization but we should probably do that separately to make this diff smaller.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -5983,9 +6006,10 @@ static void VULKAN_INTERNAL_ResetDescriptorSetData(VulkanRenderer *renderer)
 
 static void VULKAN_INTERNAL_SwapChainBlit(
 	VulkanRenderer *renderer,
-	FNA3D_Rect *sourceRectangle,
+	VulkanSwapchainData *swapchainData,
+	FNA3D_Rect * sourceRectangle,
 	FNA3D_Rect *destinationRectangle,
-	uint32_t swapChainImageIndex
+	uint32_t swapchainImageIndex
```
flibitijibibo, line 6012: Nit: Same thing with the capitalization...

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6130,17 +6158,14 @@ static void VULKAN_INTERNAL_SubmitCommands(
...
-
 	/* Must end render pass before ending command buffer */
 	VULKAN_INTERNAL_MaybeEndRenderPass(renderer, 0);
 
 	if (present)
 	{
 		SDL_GetCurrentDisplayMode(
 			SDL_GetWindowDisplayIndex(
-				renderer->presentOverrideWindowHandle
+				windowHandle
```
flibitijibibo, line 6168: This might need a cast... surprised it didn't need one before, actually.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6361,12 +6412,12 @@ static void VULKAN_INTERNAL_SubmitCommands(
 	renderer->activeCommandBufferCount = 0;
 
 	/* Present, if applicable */
-	if (present && acquireSuccess)
+	if (present && validSwapchainExists && acquireSuccess)
```
flibitijibibo, line 6415: Doesn't `!validSwapchainExists` imply `!acquireSuccess`?

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6490,256 +6544,357 @@ static void VULKAN_INTERNAL_FlushCommandsAndPresent(
...
-			&surfaceFormat
+			swapchainData->swapchainFormat,
+			swapchainSupportDetails.formats,
+			swapchainSupportDetails.formatsLength,
+			&swapchainData->surfaceFormat
 		)) {
-			SDL_free(swapChainSupportDetails.formats);
-			SDL_free(swapChainSupportDetails.presentModes);
 			FNA3D_LogError("Device does not support swap chain format");
+			if (swapchainSupportDetails.formatsLength > 0)
```
flibitijibibo, line 6627: LogError might throw, so these frees may need to be above that.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -6490,256 +6544,357 @@ static void VULKAN_INTERNAL_FlushCommandsAndPresent(
...
+			drawableHeight = clamp(
+				drawableHeight,
+				swapchainSupportDetails.capabilities.minImageExtent.height,
+				swapchainSupportDetails.capabilities.maxImageExtent.height
+			);
+		}
+		else
+		{
+			FNA3D_LogError("No fallback swapchain size available!");
+			if (swapchainSupportDetails.formatsLength > 0)
```
flibitijibibo, line 6701: Same thing as the previous comment, LogError should be after frees.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -12292,6 +12451,20 @@ static FNA3D_Device* VULKAN_CreateDevice(
 		return NULL;
 	}
 
+	/*
+	 * Create initial swapchain
+	 */
+
+	renderer->swapchainDataCapacity = 4;
```
flibitijibibo, line 12458: This can be 1 by default, 99.99% of uses will be single-window.

flibitijibibo (commented):
One final pass to be as thorough as possible, these are all pretty isolated so once these are done I'll merge it in right away. Definitely an improvement even for single-window setups!

src/FNA3D_Driver_Vulkan.c
thatcosmonaut: This gets checked when we do the dummy surface in PrepareWindowAttributes, but we have to check it for every surface we create or Vulkan validation yells at us.

src/FNA3D_Driver_Vulkan.c
thatcosmonaut: Won't this make the spelling inconsistent in the patch though?

thatcosmonaut:
Put in all the suggestions except for the capitalization thing. If you're fine with the capitalization being inconsistent in the patch then I'll add it in.

thatcosmonaut:
Found one more small thing when testing Flotilla, fixing now...

thatcosmonaut:
Alright, took care of that. Looks like Flotilla is losing track of a few VkBuffers on exit, is that expected?

flibitijibibo:
Like FEZ it's pretty leaky, so nothing to worry about there.

PR #136, opened 2022-01-05
Vulkan: Buffer synchronization rework

thatcosmonaut:
Currently, the Vulkan renderer does direct copies for buffer data uploads. We have a lot of complex logic to make sure data is synchronized correctly and that buffer data doesn't get clobbered while it's in-flight. 

This patch changes our buffer upload logic to streamline the buffer upload process. 

If NONE or DISCARD is set, we first copy the data to a staging buffer. Then we record a `vkCmdCopyBuffer` command, which is GPU-synchronized. This prevents data from being overwritten in the middle of a graphics operation. 

If NOOVERWRITE is set then we just directly copy over data like we were doing before. 

The benefits of this approach are most obvious in the `VULKAN_INTERNAL_SetBufferData` function, which is much cleaner and simpler to understand now. I was able to remove a lot of record keeping related to buffers, since they no longer need to keep track of bound sub-buffers to avoid unwanted data overwrites.

I also split the staging buffer into two separate buffers to avoid needing to stall when data is uploaded over multiple frames.

We'll probably want to test this on a bunch of traces but my preliminary tests show no problems with this patch. The only downside is an extra 64MB of allocation for the fast staging buffer.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -4862,7 +4862,7 @@ static VulkanBuffer* VULKAN_INTERNAL_CreateBuffer(
 
 	buffer->size = size;
 	buffer->resourceAccessType = resourceAccessType;
-	buffer->usage = usage | VK_BUFFER_USAGE_TRANSFER_SRC_BIT;
+	buffer->usage = usage | VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT;
 	buffer->preferDeviceLocal = isDeviceLocal;
 	buffer->isStagingBuffer = isStagingBuffer;
 
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -4959,6 +4959,7 @@ static void VULKAN_INTERNAL_CreateFastStagingBuffers(
 		1,
 		1
 	);
+	renderer->stagingBuffers[0].fastBufferOffset = 0;
 
 	renderer->stagingBuffers[1].fastBuffer = (VulkanBuffer*) VULKAN_INTERNAL_CreateBuffer(
 		renderer,
@@ -4968,13 +4969,15 @@ static void VULKAN_INTERNAL_CreateFastStagingBuffers(
 		1,
 		1
 	);
+	renderer->stagingBuffers[0].fastBufferOffset = 0;
 }
 
-static void VULKAN_INTERNAL_CreateSlowStagingBuffers(
+static void VULKAN_INTERNAL_CreateSlowStagingBuffer(
 	VulkanRenderer *renderer,
-	VkDeviceSize size
+	VkDeviceSize size,
+	uint32_t index
 ) {
-	renderer->stagingBuffers[0].slowBuffer = (VulkanBuffer*) VULKAN_INTERNAL_CreateBuffer(
+	renderer->stagingBuffers[index].slowBuffer = (VulkanBuffer*) VULKAN_INTERNAL_CreateBuffer(
 		renderer,
 		size,
 		RESOURCE_ACCESS_MEMORY_TRANSFER_READ_WRITE,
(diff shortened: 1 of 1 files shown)


flibitijibibo:
I'm a little worried that None and Discard are now the same - I can definitely see how None is better but the fact that Discard ends render passes seems like it could be a huge penalty if updates happen between target changes.

I would compare performance with SpriteBatchTest, modified to use 4x the max batch size and with SupportsNoOverwrite set to 0. Immediate will benchmark None and Deferred will benchmark Discard.

thatcosmonaut:
Confirmed that Discard is breaking the batch - SpriteBatchTest with 8192 sprites uses 4 render passes instead of 1. This does not appear to have any measurable effect on performance in this test however. We might need to look at another test case to see if this will be a problem.

What was interesting was None performance. ~7000ms on master, ~200ms with this patch.

flibitijibibo:
A test case for broken passes might actually be Apple targets - my guess is Intel and iOS will be really unhappy when you have to store and load the attachments each time you do a pass (it was definitely a problem when we were still using OpenGL on iOS). The thing that will make it ugly is less the command stream and more the Store/Load operations that happen with each pass.

thatcosmonaut:
That makes sense. I very much want to get rid of the sub-buffer system so I'm trying to think of an option for Discard - I think XNA just handed you a completely different internal buffer if you used it, so maybe we could try inserting a layer of pointer indirection. The FNA3D_Buffer handle we return is a pointer to a VulkanBuffer pointer. When Discard is used we allocate a new buffer, overwrite the pointer, mark the old buffer to be destroyed, and then we copy directly like with None.

thatcosmonaut:
Reworked the patch for a more performant DISCARD utilizing the above-mentioned strategy. We have to keep track of bound buffers again but it's much less complex than the old system.

thatcosmonaut:
Wondering if resetting `bound` is even worth it. It's a nontrivial drain on `VULKAN_INTERNAL_DestroyBuffer` and extra record keeping, and it's only efficient in the case where Discard is used every other frame on the same buffer.

thatcosmonaut:
Keeping track of `bound` is unequivocally slower in all my test cases so I'm removing it. 

That said, destroying buffers on Discard is a drain on the memory management system and creates fragmentation. I think a potential solution here to have VulkanBufferContainers hold on to `n` discarded buffers and rotate them as necessary.

flibitijibibo:
That makes sense to me - it avoids overallocating for static buffers and should level out quickly for buffers that discard frequently.

thatcosmonaut:
Alright, VulkanBufferContainers now hang on to discarded buffers and rotate them similar to the old sub-buffer system. In my tests performance seems slightly improved for Discard and dramatically improved for None.

thatcosmonaut:
Any other requests for this patch or tests we'd like to run?

flibitijibibo:
I think that covers it for me - if it passes the trace suite this is good to squash in.

thatcosmonaut:
Alright I'll run the trace suite today.

thatcosmonaut:
Rebased and ran the trace suite, looking good. Let me know if there's anything left to do here.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -4929,169 +4898,109 @@ static uint8_t VULKAN_INTERNAL_AllocateSubBuffer(
...
+
+	renderer->stagingBuffers[1].fastBuffer = (VulkanBuffer*) VULKAN_INTERNAL_CreateBuffer(
+		renderer,
+		size,
+		RESOURCE_ACCESS_MEMORY_TRANSFER_READ_WRITE,
+		VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT,
 		1,
 		1
 	);
+	renderer->stagingBuffers[0].fastBufferOffset = 0;
```
flibitijibibo, line 4972: Typo?
```suggestion
	renderer->stagingBuffers[1].fastBufferOffset = 0;
```

src/FNA3D_Driver_Vulkan.c
```diff
@@ -11264,7 +11163,6 @@ static void VULKAN_SetVertexBufferData(
 	int32_t vertexStride,
 	FNA3D_SetDataOptions options
 ) {
-	/* FIXME: use staging buffer for elementSizeInBytes < vertexStride */
```
flibitijibibo, line 11267: This FIXME is actually related to a weird thing in XNA where SetData sends data that's meant to be copied in segments rather than as a continuous block of memory, so it should be kept.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -7075,29 +6910,25 @@ static void VULKAN_INTERNAL_RemoveBuffer(
...
 	VulkanRenderer *renderer,
-	VulkanBuffer *buf
+	VulkanBuffer *vulkanBuffer
 ) {
-	VulkanSubBuffer *subbuf = buf->subBuffers[buf->currentSubBufferIndex];
-	subbuf->bound = renderer->submitCounter;
-
-	/* Don't rebind a bound buffer */
-	if (buf->bound) return;
+	if (vulkanBuffer->bound) { return; }
```
flibitijibibo, line 6917:
```suggestion
	if (vulkanBuffer->bound)
	{
		return;
	}
```

flibitijibibo (commented):
Couple small things along with the typo I found earlier, when those are in I'll squash this into master.

thatcosmonaut:
Done!

src/FNA3D_Driver_Vulkan.c
```diff
@@ -11371,6 +11274,7 @@ static void VULKAN_SetIndexBufferData(
 	int32_t dataLength,
 	FNA3D_SetDataOptions options
 ) {
+	/* FIXME: use staging buffer for elementSizeInBytes < vertexStride */
```
flibitijibibo, line 11277: Looks like it was added to the wrong one...

thatcosmonaut: Hahahaha whoops, one sec

thatcosmonaut:
Theeere we go

flibitijibibo (approved):
Okay, this is good to go. Squash this via GitHub and write a new commit message for the final changes!

Issue #140, opened 2022-01-13
Latest FNA3D dll crashes on graphics device creation and destruction 


YTN0:
Running the latest DLL, FNA3D crashes with the following message on graphics device creation (in D3D11 mode):

Exception thrown: read access violation.
output was nullptr.

Using a debug version of the DLL, Visual Studio is pointing to the following:
    FNA3D.dll!D3D11_PLATFORM_CreateSwapChain(D3D11Renderer * renderer, FNA3D_PresentationParameters * pp) Line 5412    C

It appears to be this block of code:

    IDXGIOutput_FindClosestMatchingMode(
        output,
        &swapchainBufferDesc,
        &swapchainDesc.BufferDesc,
        (IUnknown*) renderer->device
    );

This problem is occurring on my laptop which has dual GPUs viz. an integrated Intel UHD 620 and an MX150. The graphics driver is set to auto select. If I force the MX150 driver, then things work. If I leave it on auto-select or pick the UHD 620, it crashes.

FYI, the same DLL appears to works on my Dell with only an UHD 620 GPU, so I suspect this issue is somehow related to multi-gpus and picking an adapter.

If I change to using Vulkan, launching the app works fine, however, it crashes on exit when calling	FNA3D.FNA3D_DestroyDevice(GLDevice);

I get an "access violation reading location xyz" error, which is pointing to this line: 
    FNA3D.dll!ShaderResources_DeactivateDescriptorSet(ShaderResources * shaderResources, unsigned int index) Line 5667	C

Note, the Vulkan crash on exit occurs on both the single GPU laptop and the dual GPU laptop.

flibitijibibo:
@bartwe

bartwe:
Do you perhaps know if you have different directx11 feature_level support per gpu ?

flibitijibibo:
Oh also @thatcosmonaut will care about the Vulkan part

bartwe:
Ok managed to reproduce.

Two issues, the main one is that while it finds the right gpu, enumerating the outputs on that gpu doesn't produce any as those are connected to the low power gpu normally.
the other is that it uses D3D_IID_IDXGIAdapter instead of D3D_IID_IDXGIAdapter1 which i have a patch for.

The first issue i'm asking for more information from devs with more experience

bartwe:
Ok gathered more information, and they recommend calling MonitorFromWindow , finding the IDXGIOutput from all of them with that HWINDOW and using that one.

[...]

PR #141, opened 2022-01-13
Resolve DXGI_MODE_DESC on any adapter output that holds our window.

bartwe:
Related to issue #140

uses the output that contains our whnd, or the primary monitor as the source of the information used to create the swapchain.

Also resolves an issue with mixing IDXGIAdapter1 and IDXGIAdapter


src/FNA3D_Driver_D3D11.c
```diff
@@ -5379,18 +5379,55 @@ static void D3D11_PLATFORM_GetDefaultAdapter(
...
 ) {
-	IDXGIOutput *output;
 	IDXGIFactory1* pParent;
 	DXGI_SWAP_CHAIN_DESC swapchainDesc;
 	DXGI_MODE_DESC swapchainBufferDesc;
 	SDL_SysWMinfo info;
 	HWND dxgiHandle;
 	HRESULT res;
 
+
```
flibitijibibo, line 5430:
```suggestion
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -5404,17 +5441,7 @@ static void D3D11_PLATFORM_CreateSwapChain(
...
-		0,
-		&output
-	);
-	IDXGIOutput_FindClosestMatchingMode(
-		output,
-		&swapchainBufferDesc,
-		&swapchainDesc.BufferDesc,
-		(IUnknown*) renderer->device
-	);
+	ResolveSwapChainModeDescription((IUnknown*) renderer->device, (IDXGIAdapter*) renderer->adapter, (IDXGIFactory1*) renderer->factory, dxgiHandle, &swapchainBufferDesc, &swapchainDesc.BufferDesc);
```
flibitijibibo, line 5444:
```suggestion
	ResolveSwapChainModeDescription(
		(IUnknown*) renderer->device,
		(IDXGIAdapter*) renderer->adapter,
		(IDXGIFactory1*) renderer->factory,
		dxgiHandle,
		&swapchainBufferDesc,
		&swapchainDesc.BufferDesc
	);
```

flibitijibibo (approved):
Just whitespace, I'll take care of those...

bartwe:
I am wondering just slightly if we do not want to default to the original behavior, as picking the monitor that starts the application is about as arbitrary a source of settings as is the first monitor on the adapter..

flibitijibibo:
Hypothetically we could pull the window's active monitor via SDL_syswm or something, but this does seem more precise and in most cases will end up on the same display anyhow.

In any case, this is now in fnalibs.

PR #142, opened 2022-01-18
D3D11: Fall back to feature level 11_0 if 11_1 is unsupported

thatcosmonaut:
Apparently feature level 11_1 is special and will throw an error if unsupported, unlike every other feature level. Fun!

Marking as a draft until the person who reported this is able to test.

### DIFF
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -4740,6 +4740,23 @@ static uint8_t D3D11_PrepareWindowAttributes(uint32_t *flags)
 		NULL
 	);
 
+	if (FAILED(res))
+	{
+		FNA3D_LogWarn("Creating device with feature level 11_1 failed. Lowering feature level.", res);
+		res = D3D11CreateDeviceFunc(
+			NULL,
+			D3D_DRIVER_TYPE_HARDWARE,
+			NULL,
+			D3D11_CREATE_DEVICE_BGRA_SUPPORT,
+			&levels[1],
+			SDL_arraysize(levels) - 1,
+			D3D11_SDK_VERSION,
+			NULL,
+			NULL,
+			NULL
+		);
+	}
+
 	D3D11_PLATFORM_UnloadD3D11(module);
 
 	if (FAILED(res))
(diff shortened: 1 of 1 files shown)


thatcosmonaut:
The device creation fallback worked, but then crashed on creating the swapchain in Windows 7. Gonna see if specifying the adapter on device creation fixes it. 

Additionally, the debug bitflag manipulation was completely wrong so I fixed that.

src/FNA3D_Driver_D3D11.c
```diff
@@ -5014,26 +5032,39 @@ static FNA3D_Device* D3D11_CreateDevice(
...
-		/* Creating a debug mode device will fail on some systems due to the necessary
-		 * debug infrastructure not being available. Remove the debug flag and retry. */
-		FNA3D_LogWarn("Creating device in debug mode failed with error %08X. Trying non-debug.", res);
-		flags ^= D3D11_CREATE_DEVICE_DEBUG;
-		debugMode = 0;
-		goto try_create_device;
+		if (flags & D3D11_CREATE_DEVICE_DEBUG)
+		{
+			/* Creating a debug mode device will fail on some systems due to the necessary
+			* debug infrastructure not being available. Remove the debug flag and retry. */
```
flibitijibibo, line 5052:
```suggestion
			 * debug infrastructure not being available. Remove the debug flag and retry.
			 */
```

flibitijibibo (approved):
Just a bit of whitespace, this is clear to merge.

The gotos actually turned out really nice, but I guess it couldn't hurt to have a quick comment saying "this is what the CreateDevice call order looks like in the worst case". Not perfect but the alternative is to have like 5 different calls that are all wrapped around the right side of the screen.

thatcosmonaut:
I actually really dislike the gotos because reasoning about this code scrambles my brain, so I wrote in a verbose fallback process. Take it or leave it, if you don't like it I'll just revert the commit.

flibitijibibo:
Let's revert it for now, I'll try and compress this a different way.

thatcosmonaut:
Alright, done.

flibitijibibo:
Recompressed version: https://github.com/FNA-XNA/FNA3D/commit/b3525f35aef595198ae86333177a65e397f22f10

So there's still a goto, but the retries aren't interleaved anymore.

flibitijibibo:
This is now in fnalibs.

PR #144, opened 2022-01-27
Swizzle fixes

thatcosmonaut:
Really weird that you just committed a swizzle fix, was discussing this exact problem with a user in the Discord trying to get Bastion to run on RaspPi.

Internal texture formats mismatching is triggering undefined GL behavior. This patch fixes that. We also missed an incorrect swizzle on BGRA4444 in Vulkan.

### DIFF
--- a/src/FNA3D_Driver_OpenGL.c
+++ b/src/FNA3D_Driver_OpenGL.c
@@ -349,8 +349,8 @@ static int32_t XNAToGL_TextureInternalFormat[] =
 {
 	GL_RGBA8,				/* SurfaceFormat.Color */
 	GL_RGB8,				/* SurfaceFormat.Bgr565 */
-	GL_RGB5_A1,				/* SurfaceFormat.Bgra5551 */
-	GL_RGBA4,				/* SurfaceFormat.Bgra4444 */
+	GL_BGRA,				/* SurfaceFormat.Bgra5551 */
+	GL_BGRA,				/* SurfaceFormat.Bgra4444 */
 	GL_COMPRESSED_RGBA_S3TC_DXT1_EXT,	/* SurfaceFormat.Dxt1 */
 	GL_COMPRESSED_RGBA_S3TC_DXT3_EXT,	/* SurfaceFormat.Dxt3 */
 	GL_COMPRESSED_RGBA_S3TC_DXT5_EXT,	/* SurfaceFormat.Dxt5 */
@@ -367,7 +367,7 @@ static int32_t XNAToGL_TextureInternalFormat[] =
 	GL_RG16F,				/* SurfaceFormat.HalfVector2 */
 	GL_RGBA16F,				/* SurfaceFormat.HalfVector4 */
 	GL_RGBA16F,				/* SurfaceFormat.HdrBlendable */
-	GL_RGBA8,				/* SurfaceFormat.ColorBgraEXT */
+	GL_BGRA,				/* SurfaceFormat.ColorBgraEXT */
 	GL_SRGB_ALPHA_EXT,			/* SurfaceFormat.ColorSrgbEXT */
 	GL_COMPRESSED_SRGB_ALPHA_S3TC_DXT5_EXT,	/* SurfaceFormat.Dxt5SrgbEXT */
 	GL_COMPRESSED_RGBA_BPTC_UNORM_EXT,	/* SurfaceFormat.BC7EXT */


flibitijibibo:
I ended up finding this one during Wayland testing this week. Should be back in the Discord eventually!

src/FNA3D_Driver_OpenGL.c
```diff
@@ -367,7 +367,7 @@ static int32_t XNAToGL_TextureInternalFormat[] =
 	GL_RG16F,				/* SurfaceFormat.HalfVector2 */
 	GL_RGBA16F,				/* SurfaceFormat.HalfVector4 */
 	GL_RGBA16F,				/* SurfaceFormat.HdrBlendable */
-	GL_RGBA8,				/* SurfaceFormat.ColorBgraEXT */
+	GL_BGRA,				/* SurfaceFormat.ColorBgraEXT */
```
flibitijibibo, line 370: This is a spec violation to my knowledge - it's supposed to be BGRA with RGBA8 as the storage format; AFAIK this is the closest we can get to compliance since ES3 doesn't support BGRA order.

src/FNA3D_Driver_OpenGL.c
```diff
@@ -349,8 +349,8 @@ static int32_t XNAToGL_TextureInternalFormat[] =
 {
 	GL_RGBA8,				/* SurfaceFormat.Color */
 	GL_RGB8,				/* SurfaceFormat.Bgr565 */
-	GL_RGB5_A1,				/* SurfaceFormat.Bgra5551 */
-	GL_RGBA4,				/* SurfaceFormat.Bgra4444 */
+	GL_BGRA,				/* SurfaceFormat.Bgra5551 */
```
flibitijibibo, line 352: These two I for sure need XNA comparison tests. I used unit tests for most of these subbyte formats years ago and haven't touched them since.

src/FNA3D_Driver_Vulkan.c
```diff
@@ -1536,12 +1536,7 @@ static VkComponentMapping XNAToVK_SurfaceSwizzle[] =
 		VK_COMPONENT_SWIZZLE_R,
 		VK_COMPONENT_SWIZZLE_A
 	},
-	{			/* SurfaceFormat.Bgra4444 */
-		VK_COMPONENT_SWIZZLE_B,
-		VK_COMPONENT_SWIZZLE_G,
-		VK_COMPONENT_SWIZZLE_R,
-		VK_COMPONENT_SWIZZLE_A
-	},
+	IDENTITY_SWIZZLE,	/* SurfaceFormat.Bgra4444 */
```
flibitijibibo, line 1539: This is fine, can be committed separately if you want

thatcosmonaut:
Hold off on merging this, investigating a GL crash.

flibitijibibo (changes requested):
Vulkan fix is fine, GL changes I'm almost positive are all wrong

thatcosmonaut:
Alright I'll continue investigating. Will commit the Vulkan change separately

flibitijibibo:
Note that the array being touched here is bit size and not swizzling, that's done in the array right above this one.

If this was reported from an ES3 user this is not a bug, it's just ES3 omitting BGRA8 for some unknown reason.

thatcosmonaut:
Actually the BGRA error is on mainline OpenGL and not ES3. Trying to get to the bottom of this. There's a depthRange error on Nvidia GL as well that I'm looking into.

flibitijibibo:
This is all very very highly suspicious - ARB_debug_output would have been screaming about this 6 years ago. NV rightly errors on BGRA in ES3 but there's no way it trips in desktop GL. Depth range it _might_ claim that double precision is unsupported in ES3 but this isn't a hard error (and on Tegra you can either ignore it or force a desktop profile).

I don't know what's going on but the fact that we're suddenly getting a report for GL formats reeks of someone secretly using some crap like a mobile vendor's driver.

thatcosmonaut:
According to this user both desktop GL and ES3 are crashing on Intel HD 630 using Mesa 21.3.4

flibitijibibo:
Why are we even testing ES3 on Mesa??? That's not going to help at all and never has.

If it's crashing with desktop GL that needs to be reported to the Mesa team, not us:

https://docs.mesa3d.org/bugs.html

JohnnyonFlame:
The mismatch between internal format and format affects all platforms I have that are ES3 capable, NV, Intel, Mali Midgard r18p0 and Mali Bifrost r24p0. On Mali it causes texture corruptions, on NV/Intel it causes a crash.

However the proposed changes for OpenGL ES3 causes the OpenGL path to fail as reported.

flibitijibibo:
This is this third time I've said it in this thread alone and it had better be the last:

ES3 DOES NOT SUPPORT BGRA8. THIS IS HOW ES3 WORKS. IT IS NOT A BUG, IT IS JUST A SPEC OVERSIGHT.

I can only assume that because we're even looking at this you're trying to run Bastion. Don't even try it on ES hardware, it has a GetData in the current public build that basically makes it a no-go. I have this fixed locally but Supergiant haven't taken the patches in. Until they take those patches in, you're not running this game. Play something else.

thatcosmonaut:
My bad, I was under the impression this was a desktop GL issue, apologies for the time waste.

JohnnyonFlame:
> ES3 DOES NOT SUPPORT BGRA8. THIS IS HOW ES3 WORKS. IT IS NOT A BUG, IT IS JUST A SPEC OVERSIGHT.
Loud and clear, I had tested it on mesa/nv/mali on ES3 without actually checking spec then reported it on Discord, sorry for the bogus report cosmo, I really hate this API sometimes.

PR #146, opened 2022-02-01
Vulkan: Fix SetBufferData race on defrag frames

thatcosmonaut:
The defragmentation process performs async copies on the GPU, which we don't control the timing of, so under certain conditions the GPU can perform the copy after SetBufferData is called which will destroy user data. For None, async copies are performed using memory barriers so no race is possible. For Discard, marking the moved buffer as bound is enough to prevent a race because an unbound buffer will be selected or created. For NoOverwrite we have no choice but to wait on the defrag fence.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -1895,8 +1895,8 @@ static void VULKAN_INTERNAL_ImageMemoryBarrier(
 );
 
 static void VULKAN_INTERNAL_MarkBufferAsBound(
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):

thatcosmonaut:
Sanity checking with the SOR4 mega trace before merging.

thatcosmonaut:
Had to move MarkBufferASBound to forward declarations but the performance on the mega trace is looking great.

flibitijibibo (approved):
Still looks fine, if the traces are happy this can go in.

PR #148, opened 2022-02-14
Vulkan: Submit synchronization rewrite

thatcosmonaut:
Multi-window exposed a lot of flawed assumptions in how we are dealing with submission flow. With single window, we could assume that submission was a good time to wait and clean up and rotate inactive resources before proceeding with submission. Now that multiple windows can be in flight at once, this assumption means that we are waiting on GPU work for no actual reason. It also meant we had to have two levels of destroyed resource storage - one for resources that are still in flight, and then one for resources that are actually ready to be destroyed.

I realized that rather than having bound resources live in the renderer context, it's better to keep track of them per command buffer. Now we have a thread wait on the GPU work to be completed before cleaning up the resources that were bound for that set of command buffer work, so only one level of storage is required.

Multi-window also showed that our descriptor set management was flawed. We were keeping track of a cache of sampler descriptor sets and invalidating them after X frames, but this assumption kind of falls apart when multiple windows are being submitted. There also wasn't any real performance benefit to caching, since I have now discovered that it is not descriptor set *writes* that are expensive, but descriptor set *allocations*. We still have a pool of sampler descriptor sets per shader, but now when the sampler texture set changes we just grab an inactive one and update it with the current sampler texture set. This means we don't have to keep track of a huge cache of descriptor sets and we also don't have to worry about invalidating individual descriptor sets on defragmentation. 

This is a very large change to a fundamental component of the renderer, so we will need to run the usual tests and the full trace set, but I'm submitting this now in case there are any objections to the structure of the changes.


src/FNA3D_Driver_Vulkan.c
```diff
@@ -2965,6 +2997,7 @@ static void VULKAN_INTERNAL_MakeMemoryUnavailable(
 
 				allocation->allocator->sortedFreeRegions[j]->sortedIndex = j;
 			}
+
```
flibitijibibo, line 3000:
```suggestion
```

flibitijibibo (commented):
Just one extra line from what I could tell!

thatcosmonaut:
We had a chat in the Discord about threading vs polling and I decided that polling for command buffer completion is a superior strategy over waiting in a thread. `vkWaitForFences` apparently lets you query command buffer state when you set a timeout value of 0, so now in SwapBuffers we poll all submitted command buffers for completion. If a submitted CB is complete, we perform cleanup tasks. 

I renamed staging buffers to transfer buffers, and these are now pooled and assigned to specific command buffers. If a command buffer used a transfer buffer, it is returned to the pool on cleanup. This lets us avoid odd race conditions or having to stall when multiple frames are in flight.

Dummy textures are now marked as dedicated to avoid invalidating dummy descriptor sets on defragmentation.

flibitijibibo (approved):
Nothing stood out as wrong, but we're looking at a 2600-line diff. Valgrind with leak checks and Vulkan validation should be run on all existing traces before merging.

thatcosmonaut:
I've been running traces with validation on and I ran valgrind when testing, but I'll run the traces through vg once more just to be extra careful.

thatcosmonaut:
Ran the traces, no validation errors or leaks reported.

PR #153, opened 2022-09-11
Add missing render pass mutex unlock

8212369:
I think it should be unlocked afterwards, I was experiencing some freeze with this mutex and found out that at this call it was not unlocked. Add an unlock fixed the freeze.


flibitijibibo:
I believe this is meant to be a long-term lock that gets unlocked in EndRenderPass but @thatcosmonaut will know better than me... we actually seem to unlock twice in a row so something seems unbalanced here.

8212369:
End render pass lock the mutex again, so I don't think we unlock it twice. But I will wait!

8212369:
Sorry, this was my mistake， it indeed is locking /unlocking right pair. I will investigate what's wrong on my local. It was stucking at ApplyEffect function and this mutex for more context.

PR #154, opened 2022-09-11
Vulkan: Use identity transform for Vulkan swapchain

8212369:
XNA game works with rotated framebuffer, so Android has to rotate them again for native display (for [reference](https://community.arm.com/arm-community-blogs/b/graphics-gaming-and-vr-blog/posts/appropriate-use-of-surface-rotation)). So I use identity transform to fix the graphics.

Without the change I got some ported games squeezed in potrait resolution while it's meant to be in landscape.


8212369:
I don't think this variable is used for desktop, so it should not be a problem on Desktop.

flibitijibibo:
Would this be something that DisplayOrientation cares about? https://github.com/FNA-XNA/FNA3D/blob/master/include/FNA3D.h#L76

Patch makes sense to me, but wanted to make sure that we can't expose the driver's value somehow.

flibitijibibo:
I read the spec more closely and you're correct, in order for us to use this we have to rotate our swapchain blit even if we have chain images using the right dimensions... which seems wrong to me, but I usually think in OpenGL terms where we aren't making the allocations.

Thanks for the obscure find!

Issue #165, opened 2023-04-07
Excessive render pass restarts for each draw


flibitijibibo:
Pretty sure this is https://github.com/FNA-XNA/FNA3D/commit/b1c55d70d10a16de432e906e35c920564206adb7 acting up - basically all you need to do is run a game with Vulkan and inspect a RenderDoc capture; FEZ in particular showed an alarming number of render passes and our barrier system ending render passes appears to be the culprit. We'll probably need to revert this commit and instead add another check to ensure that there's a _reason_ for the barrier (not just the layout, but memory activity as well).

thatcosmonaut:
Resolved by https://github.com/FNA-XNA/FNA3D/pull/166

PR #166, opened 2023-04-07
Vulkan: Restructure memory barriers to avoid breaking render pass

thatcosmonaut:
Addresses https://github.com/FNA-XNA/FNA3D/issues/165

We should really only have been setting memory barriers in cases where a data race was actually possible, so this patch addresses that. Memory barriers are only created on data copies, and when a render pass ends.

To help facilitate this change, SetRenderTargets now configures the state for the next render pass. That state is then set when the new render pass is triggered, instead of immediately.

The validation layer might complain now if you bind a resource that hasn't had any data written to it, but that seems acceptable to me.


flibitijibibo:
Haven't identified the issue but [SpriteBatchTest](https://gist.github.com/flibitijibibo/27c486c236301b49b650) is a black screen with this patch.

PR #171, opened 2023-06-18
[WIP] Refactor Vulkan's memory management into common code

TheSpydog:
Once complete, this should help future backends (and SDL_GPU) wrangle memory management without having to copy-paste a bunch of logic from the Vulkan driver.

Rough Todo List:
- [ ] Resource memory binding
- [ ] Defrag
- [ ] Testing
- [ ] Less verbose naming?
- [ ] Other build systems (visualc-gdk, Xcode-iOS, etc)


TheSpydog:
I think this is going in the wrong direction design-wise. Closing for now, will try something else.

PR #172, opened 2023-06-23
Shared Memory Manager

flibitijibibo:
This diff as a whole looks horrible, so definitely use the commit history instead!

Mostly looking for two types of reviews:
- @thatcosmonaut, make sure I didn't explode the Vulkan renderer in the process >_>
- @TheSpydog, give this draft a try and see if we need to make changes to accommodate the other renderers

Known Issues:
- [ ] Track device-local max heap use in the allocator?
- [ ] Need to remove dedicated allocation from Vulkan side
- [ ] The void* is just directly cast to VkDeviceMemory, which is kinda bad
- [ ] VULKAN_Memory_CreateBufferHandle can be simplified (IMO)
- [ ] VULKAN_Memory_MarkBufferHandlesForDestroy needs to include the whole array passed to it


src/FNA3D_Memory.h
```diff
@@ -0,0 +1,253 @@
...
+#include <SDL.h>
+
+/* Driver-defined memory structures  */
+
+typedef struct FNA3D_MemoryHandle FNA3D_MemoryHandle;
+typedef struct FNA3D_BufferHandle FNA3D_BufferHandle;
+
+/* Resource Access Types are mostly for setting up pipeline barriers */
+
+typedef enum FNA3D_Memory_ResourceAccessType
```
flibitijibibo, line 39: This can probably go back in Vulkan.c; I thought it would be used more but really we just need to have a vertex/index buffer bool in CreateBufferContainer (and the associated driver function).

thatcosmonaut: I'm actually tempted to go the other direction - the renderer does need to indicate when barriers are needed, but I don't think it should necessarily be keeping track of the barrier state itself

flibitijibibo: For now I've reverted this part, but if we need to expose it we can - this was only used to initialize vertex/index buffers; the barriers are handled within the Vulkan renderer without the allocator's involvement.

flibitijibibo:
All of the FIXMEs except for the VkDeviceMemory casting have been resolved - miraculously this fixed most of the ugliness of the git diff as well, so once we know this is okay I'll squash this in. (Or at least, I'll squash all the commits up until now, would be nice to fix the flicker we found in the Wizorb trace.)

PR #177, opened 2023-09-29
Add SetEnv and GetEnv APIs for overriding FNA3D env vars at runtime

Chicken-Bones:
Required for https://github.com/FNA-XNA/FNA/pull/437

### Discussion
- API naming
  - Use SDL `Hint` naming?
  - Full names? `Set/GetEnvironmentVariable`
  - Put `Override` in the name?
- Documentation quality
  - The use of 'may' to leave for implementation flexibility may (heh) cause confusion. COnsider simpler language which describes the current implementation rather than just the API surface.

### DIFF
--- a/include/FNA3D.h
+++ b/include/FNA3D.h
@@ -493,6 +493,20 @@ FNA3DAPI void FNA3D_HookLogFunctions(
 	FNA3D_LogFunc error
 );
 
+/* Environment */
+
+/* Gets the value of an environment variable.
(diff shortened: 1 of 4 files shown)


flibitijibibo:
We won't need to override this in FNA3D, since we only use hints and don't interact with the CLR anyhow. Fixing it in FNA.dll will cover all the bug scenarios in this case.

PR #178, opened 2023-10-01
Add support for using WARP software rasterizer with the D3D11 backend

kg:
by setting FNA3D_D3D11_USE_WARP=1

WARP software rasterizer ships with Windows and is fast enough on multicore machines to be useful as a debugging tool (to rule out driver issues, etc) or even as a manual fallback if the user has a bad GPU.

See https://learn.microsoft.com/en-us/windows/win32/direct3darticles/directx-warp and https://learn.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-devices-create-warp

### DIFF
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -4848,8 +4848,7 @@ static uint8_t D3D11_PrepareWindowAttributes(uint32_t *flags)
 	};
 	HRESULT res;
 
-	const char* useWarp = SDL_GetHint("FNA3D_D3D11_USE_WARP");
-	const uint32_t driverType = (useWarp == NULL) || (SDL_strcmp(useWarp, "1") != 0)
+	const uint32_t driverType = SDL_GetHintBoolean("FNA3D_D3D11_USE_WARP", SDL_FALSE)
 		? D3D_DRIVER_TYPE_HARDWARE
 		: D3D_DRIVER_TYPE_WARP;
 
@@ -5115,8 +5114,7 @@ static FNA3D_Device* D3D11_CreateDevice(
 	int32_t i;
 	HRESULT res;
 
-	const char* useWarp = SDL_GetHint("FNA3D_D3D11_USE_WARP");
-	const uint32_t driverType = (useWarp == NULL) || (SDL_strcmp(useWarp, "1") != 0)
+	const uint32_t driverType = SDL_GetHintBoolean("FNA3D_D3D11_USE_WARP", SDL_FALSE)
 		? D3D_DRIVER_TYPE_UNKNOWN /* Must be UNKNOWN if adapter is non-null according to spec */
 		: D3D_DRIVER_TYPE_WARP;
 
@@ -5187,7 +5185,7 @@ try_create_device:
 	for (i = 0; i < 2; i += 1)
 	{
 		res = D3D11CreateDeviceFunc(
-			driverType == D3D_DRIVER_TYPE_WARP ? NULL : (IDXGIAdapter*) renderer->adapter,
+			(driverType == D3D_DRIVER_TYPE_WARP) ? NULL : (IDXGIAdapter*) renderer->adapter,
 			driverType, 
 			NULL,
 			flags,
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_D3D11.c
```diff
@@ -4848,6 +4848,11 @@ static uint8_t D3D11_PrepareWindowAttributes(uint32_t *flags)
 	};
 	HRESULT res;
 
+	const char* useWarp = SDL_GetHint("FNA3D_D3D11_USE_WARP");
+	const uint32_t driverType = (useWarp == NULL) || (SDL_strcmp(useWarp, "1") != 0)
+		? D3D_DRIVER_TYPE_HARDWARE
+		: D3D_DRIVER_TYPE_WARP;
```
flibitijibibo, line 4854:
```suggestion
	const uint32_t driverType = SDL_GetHintBoolean("FNA3D_D3D11_USE_WARP", SDL_FALSE)
		? D3D_DRIVER_TYPE_HARDWARE
		: D3D_DRIVER_TYPE_WARP;
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -5177,8 +5187,8 @@ static FNA3D_Device* D3D11_CreateDevice(
 	for (i = 0; i < 2; i += 1)
 	{
 		res = D3D11CreateDeviceFunc(
-			(IDXGIAdapter*) renderer->adapter,
-			D3D_DRIVER_TYPE_UNKNOWN, /* Must be UNKNOWN if adapter is non-null according to spec */
+			driverType == D3D_DRIVER_TYPE_WARP ? NULL : (IDXGIAdapter*) renderer->adapter,
```
flibitijibibo, line 5190:
```suggestion
			(driverType == D3D_DRIVER_TYPE_WARP) ? NULL : (IDXGIAdapter*) renderer->adapter,
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -5110,6 +5115,11 @@ static FNA3D_Device* D3D11_CreateDevice(
 	int32_t i;
 	HRESULT res;
 
+	const char* useWarp = SDL_GetHint("FNA3D_D3D11_USE_WARP");
+	const uint32_t driverType = (useWarp == NULL) || (SDL_strcmp(useWarp, "1") != 0)
+		? D3D_DRIVER_TYPE_UNKNOWN /* Must be UNKNOWN if adapter is non-null according to spec */
+		: D3D_DRIVER_TYPE_WARP;
```
flibitijibibo, line 5121:
```suggestion
	const uint32_t driverType = SDL_GetHintBoolean("FNA3D_D3D11_USE_WARP", SDL_FALSE)
		? D3D_DRIVER_TYPE_UNKNOWN /* Must be UNKNOWN if adapter is non-null according to spec */
		: D3D_DRIVER_TYPE_WARP;
```

flibitijibibo (approved):
A few changes to shrink the diffs via SDL_GetHintBoolean, otherwise fine!

Popax21:
The refactoring changes in 37ec1d9 seemed to have introduced a severe bug - see #179.

kg:
Good catch @Popax21 , I didn't notice it since I didn't do a fresh local build after applying the code review suggestions.

flibitijibibo:
Will merge the fix as soon as it's submitted!

Popax21:
> Will merge the fix as soon as it's submitted!

Well, the fix is just swapping the options around for both ternaries. Since it's such a simple fix I was under the impression that going through a PR might be more effort from the overhead, however if it's easier for you I can quickly create one.

flibitijibibo:
Go for it, I'll be in the office in a bit so I can merge and then rebuild fnalibs as soon as I'm in!

Popax21:
> Go for it, I'll be in the office in a bit so I can merge and then rebuild fnalibs as soon as I'm in!

Dang, seems like I missed the time window by just a bit .-. (context: commit fcfab15 just fixed this)

PR #183, opened 2023-11-07
Fix validation warning when rendering to mipped RT in vulkan

kg:
When preparing to render to a mipmapped render target, we need to transition all of its mip levels into color attachment state (they would otherwise be in read only state) in order to satisfy the vulkan validation layer.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -6903,7 +6903,7 @@ static void VULKAN_INTERNAL_BeginRenderPass(
 			0,
 			renderer->colorAttachments[i]->layerCount,
 			0,
-			1,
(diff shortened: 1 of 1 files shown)


flibitijibibo:
I believe this change is covered by ResolveTarget, but would have to check locally... AFAIK this is good

kg:
For context, this PR makes this set of validator errors go away:
image

thatcosmonaut:
https://github.com/FNA-XNA/FNA3D/blob/master/src/FNA3D_Driver_Vulkan.c#L5328

The real bug is that the rtView contains every mip level when it should only contain level 0. Probably also true for the layers as well.

kg:
This is the wrong fix, I think @thatcosmonaut is on it though

PR #185, opened 2023-11-07
Vulkan: Fix mipped render target barrier inconsistencies

thatcosmonaut:
Real fix for #183


kg:
This doesn't seem to fix it, maybe I did something wrong when testing it locally. Maybe something else is also trying to transition all the mips?
image

kg:
We discussed on discord that there are other problems at work explaining why I get validation errors. This change still seems correct

thatcosmonaut:
Fixes #188

flibitijibibo (approved):
Nothing stands out to me, so if this is the final draft it can be squashed in.

PR #187, opened 2023-11-08
Add FNA3D_SetTextureName debugging API

kg:
I didn't implement tracing (yet?) because I'm not completely sure how to do it. I have an FNA change to go with this that I can make a PR for later

### DIFF
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -4813,6 +4813,16 @@ static void D3D11_SetStringMarker(FNA3D_Renderer *driverData, const char *text)
 	ID3DUserDefinedAnnotation_SetMarker(renderer->annotation, wstr);
 }
 
+const GUID GUID_D3DDebugObjectName = { 0x429b8c22, 0x9188, 0x4b0c, 0x87, 0x42, 0xac, 0xb0, 0xbf, 0x85, 0xc2, 0x00 };
+
+static void D3D11_SetTextureName(FNA3D_Renderer* driverData, FNA3D_Texture* texture, const char* text)
+{
+	D3D11Renderer* renderer = (D3D11Renderer*)driverData;
+	D3D11Texture* d3dTexture = (D3D11Texture*)texture;
+
+	ID3D11DeviceChild_SetPrivateData(d3dTexture->handle, &GUID_D3DDebugObjectName, strlen(text), text);
+}
+
 /* External Interop */
 
 static void D3D11_GetSysRenderer(
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -4813,14 +4813,14 @@ static void D3D11_SetStringMarker(FNA3D_Renderer *driverData, const char *text)
 	ID3DUserDefinedAnnotation_SetMarker(renderer->annotation, wstr);
 }
 
-const GUID GUID_D3DDebugObjectName = { 0x429b8c22, 0x9188, 0x4b0c, 0x87, 0x42, 0xac, 0xb0, 0xbf, 0x85, 0xc2, 0x00 };
+static const GUID GUID_D3DDebugObjectName = { 0x429b8c22, 0x9188, 0x4b0c, 0x87, 0x42, 0xac, 0xb0, 0xbf, 0x85, 0xc2, 0x00 };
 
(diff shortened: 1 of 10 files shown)


thatcosmonaut:
This change looks good to me but I think we might want to make it slightly more generalized so that we can name other types of resources.

kg:
> This change looks good to me but I think we might want to make it slightly more generalized so that we can name other types of resources.

I agree, but I'm not sure how we would go about it. I guess it would take a void* along with an enum specifying which type of FNA object we want to assign a name to?

For D3D all of the device child objects share a base interface that contains SetPrivateData, so the code there wouldn't need to change much. For vulkan I think we would just select the appropriate objectType, probably with a table.

src/FNA3D_Driver_D3D11.c
```diff
@@ -4813,6 +4813,16 @@ static void D3D11_SetStringMarker(FNA3D_Renderer *driverData, const char *text)
...
 }
 
+const GUID GUID_D3DDebugObjectName = { 0x429b8c22, 0x9188, 0x4b0c, 0x87, 0x42, 0xac, 0xb0, 0xbf, 0x85, 0xc2, 0x00 };
+
+static void D3D11_SetTextureName(FNA3D_Renderer* driverData, FNA3D_Texture* texture, const char* text)
+{
+	D3D11Renderer* renderer = (D3D11Renderer*)driverData;
+	D3D11Texture* d3dTexture = (D3D11Texture*)texture;
+
+	ID3D11DeviceChild_SetPrivateData(d3dTexture->handle, &GUID_D3DDebugObjectName, strlen(text), text);
```
flibitijibibo, line 4823:
```suggestion
	ID3D11DeviceChild_SetPrivateData(d3dTexture->handle, &GUID_D3DDebugObjectName, SDL_strlen(text), text);
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -4813,6 +4813,16 @@ static void D3D11_SetStringMarker(FNA3D_Renderer *driverData, const char *text)
 	ID3DUserDefinedAnnotation_SetMarker(renderer->annotation, wstr);
 }
 
+const GUID GUID_D3DDebugObjectName = { 0x429b8c22, 0x9188, 0x4b0c, 0x87, 0x42, 0xac, 0xb0, 0xbf, 0x85, 0xc2, 0x00 };
```
flibitijibibo, line 4816:
```suggestion
static const GUID GUID_D3DDebugObjectName = { 0x429b8c22, 0x9188, 0x4b0c, 0x87, 0x42, 0xac, 0xb0, 0xbf, 0x85, 0xc2, 0x00 };
```

src/FNA3D_Driver_OpenGL.c
```diff
@@ -5343,6 +5343,12 @@ static void OPENGL_SetStringMarker(FNA3D_Renderer *driverData, const char *text)
 	}
 }
 
+static void OPENGL_SetTextureName(FNA3D_Renderer* driverData, FNA3D_Texture* texture, const char* text)
+{
+	// No OpenGL API for this that I'm aware of -kg
```
flibitijibibo, line 5348:
```suggestion
	/* No OpenGL API for this that I'm aware of -kg */
```

src/FNA3D_Tracing.c
```diff
@@ -1549,6 +1549,12 @@ void FNA3D_Trace_SetStringMarker(const char *text)
 	SDL_UnlockMutex(traceLock);
 }
 
+void FNA3D_Trace_SetTextureName(void *texture, const char *text)
+{
+	// FIXME: Not implemented -kg
```
flibitijibibo, line 1554:
```suggestion
	SDL_assert(0 && "Trace_SetTextureName not implemented -kg");
```

flibitijibibo (commented):
This seems okay to me at least - I wouldn't mind having separate exports for each type, especially since each type will probably have to handle things differently (especially backends with the fancy buffer memory management).

Got a few cleanups listed below, then this should be okay.

src/FNA3D_Tracing.c
kg: Won't this break tracing though? Vs omit it from traces

flibitijibibo: It'll warn when recording the trace, but it shouldn't affect the output data.

kg: In release builds it appears to do nothing, which is good enough for me

PR #190, opened 2024-01-01
FNA3D: Extend displayOrientation enum by PORTRAIT_FLIPPED.

jfmu:
Those are not handled anywhere yet, but they should have the same values as in FNA (its a bitfield there).


jfmu:
Was fixed in FNA, by handling PORTRAIT_FLIPPED as PORTRAIT.Closing this.

PR #191, opened 2024-01-04
Vulkan: Fix errors on iOS/tvOS simulator and older Apple devices

TheSpydog:
I noticed when testing on Apple simulators that our Vulkan driver would error out in two different places -- when requesting the `occlusionQueryPrecise` device feature, and when using a non-zero base vertex in `vkCmdDrawIndexed`. This patch addresses both scenarios, allowing games to be tested via simulator.

I chose to make `occlusionQueryPrecise` optional because it's such a niche feature that we can live without it on low-end machines. I don't know of a single XNA game that actually uses occlusion queries, but even if there is such a game, it's highly likely they just check if the PixelCount is > 0 (which is what [the examples in the XNA documentation](https://learn.microsoft.com/en-us/previous-versions/windows/xna/bb975835(v=xnagamestudio.40)#remarks) do). That should still work even without precise occlusion queries.

For base vertex, I followed the OpenGL driver's strategy of adjusting the offset the vertex buffer in ApplyVertexBufferBindings. This works from my testing, but this is the part I feel the least confident about. I especially don't love that it adds a couple conditionals, but I'm not sure there's a way around that.

### DIFF
--- a/src/FNA3D_Driver_Vulkan.c
+++ b/src/FNA3D_Driver_Vulkan.c
@@ -1272,6 +1272,8 @@ typedef struct VulkanRenderer
 	uint8_t supportsDebugUtils;
 	uint8_t supportsDeviceProperties2;
 	uint8_t supportsSRGBRenderTarget;
+	uint8_t supportsPreciseOcclusionQueries;
+	uint8_t supportsBaseVertex;
 	uint8_t debugMode;
 	VulkanExtensions supports;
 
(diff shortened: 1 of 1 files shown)


thatcosmonaut:
Looks good to me, the conditionals are unfortunate but I think having a functioning Apple workflow is a worthy tradeoff.

flibitijibibo (approved):
If SpriteBatch works then this should be good to go, probably worth keeping this in mind for SDL anyhow.

PR #194, opened 2024-01-20
Add FNA3D_REDUCED_FRAME_LATENCY support for D3D11

kg:
By default up to 3 frames can be queued for rendering, which increases rendering latency by default. In my testing reducing the queue length to 1 reduces windowed mode max display/render latency from 30ms/47ms to 20ms/30ms, and the average latency is improved further.

I made the name generic since it might be possible to do this for Vulkan too somehow. In my testing OpenGL already has ideal latency (much lower than D3D11 or Vulkan). Vulkan's windowed latency on my test configuration is the same as D3D11's, while OpenGL's is 16ms/16ms.

In my test scenario with this active, the maximum FPS (when vsync is off) goes down from 4000 to around 3300. I don't know if this would meaningfully negatively impact real games, but it seems like it would be really useful for games like Celeste.

Measurements from PresentMon.


kg:
For context I found some Unity games that appear to do this based on their D3D11 latency. Unreal Engine does this by default in specific cases, and has a configuration variable for it that defaults to 3. So I think it makes sense to keep this off by default if we add it.

kg:
We may also want to set swapchainDesc.BufferCount to 2 in reduced latency mode instead of the default of 3.

kg:
> We may also want to set swapchainDesc.BufferCount to 2 in reduced latency mode instead of the default of 3.

Doing this reduces the latency to 13ms/24ms on my system.

kg:
On my end by default with the SDL_GPU backend, latency with vsync off is 1 frame and latency with vsync on is 2 frames. So I don't think we need to do anything for that backend.

kg:
I can't check D3D12 since it can't compile my shaders, so someone else would have to test it.

PR #197, opened 2024-02-02
Vulkan: Adjust allocation strategy

thatcosmonaut:
It turns out that copying to the BAR from the CPU is actually slow, so the concept of a "fast transfer buffer" was not correct in the first place. Getting rid of this streamlines the transfer logic, reduces VRAM usage, and may improve performance in texture upload bottleneck scenarios.

There's also no need to specify ignored memory properties - the driver is specific about what kind of memory types the GetMemoryRequirements functions return, so there's no need to constrict that further.

Additionally, setting the CACHED bit on buffer allocation is a massive speedup.

### DIFF
--- a/src/FNA3D_CommandBuffer.c
+++ b/src/FNA3D_CommandBuffer.c
@@ -29,16 +29,12 @@
 #include <SDL.h>
 
 #define STARTING_TRANSFER_BUFFER_SIZE 8000000 /* 8MB */
-#define FAST_TRANSFER_SIZE 64000000 /* 64MB */
 
 typedef struct FNA3D_TransferBufferPool
 {
-	FNA3D_TransferBuffer *fastTransferBuffer;
-	uint8_t fastTransferBufferAvailable;
-
-	FNA3D_TransferBuffer **availableSlowTransferBuffers;
-	uint32_t availableSlowTransferBufferCount;
-	uint32_t availableSlowTransferBufferCapacity;
+	FNA3D_TransferBuffer **availableTransferBuffers;
+	uint32_t availableTransferBufferCount;
+	uint32_t availableTransferBufferCapacity;
 } FNA3D_TransferBufferPool;
 
 /* Command buffers have various resources associated with them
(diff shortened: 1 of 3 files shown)


src/FNA3D_CommandBuffer.c
```diff
@@ -225,24 +221,17 @@ static void FNA3D_CommandBuffer_CleanContainer(
 		transferBuffer = container->transferBuffers[i];
 		transferBuffer->offset = 0;
 
-		if (transferBuffer == manager->transferBufferPool.fastTransferBuffer)
+		if (manager->transferBufferPool.availableTransferBufferCount == manager->transferBufferPool.availableTransferBufferCapacity)
 		{
-			manager->transferBufferPool.fastTransferBufferAvailable = 1;
+			manager->transferBufferPool.availableTransferBufferCapacity += 1;
```
kg, line 226: i'm guessing we have so few of these that growing by 1 is fine instead of by larger amounts or a ratio?

src/FNA3D_Driver_Vulkan.c
```diff
@@ -2914,32 +2892,24 @@ static uint8_t VULKAN_INTERNAL_BindMemoryForBuffer(
...
 			VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
 
-		/* Follow-up for the warning logged by FindMemoryType */
-		if (!renderer->unifiedMemoryWarning)
-		{
-			FNA3D_LogWarn("No unified memory found, falling back to host memory");
-			renderer->unifiedMemoryWarning = 1;
-		}
-
 		while (VULKAN_INTERNAL_FindBufferMemoryRequirements(
```
kg, line 2909: what happens if we never find any suitable memory? it looks like we fall through to the bottom or we loop forever?

kg (approved):
I don't see any errors

src/FNA3D_CommandBuffer.c
thatcosmonaut: yeah, these are allocated infrequently

src/FNA3D_Driver_Vulkan.c
thatcosmonaut: memoryTypeIndex is incremented on these calls, so if an alloc fails on a particular memory type it can try again with another memory type. eventually this function will fail and cause the memory bind to fail. at that point the system would be completely out of memory so everything is extremely screwed anyway

flibitijibibo (approved):
Looks okay - since this affects CommandBuffer we should try to update PS5 at the same time.

TheSpydog:
Updated PS5 with the ABI change (it wasn’t using that parameter anyway) so this is good to go on that front.

thatcosmonaut:
Our SDL GPU proposal is approved now, and that Vulkan implementation is enormously superior to FNA3D's, so I'm closing this with the expectation that we will be deleting FNA3D_Driver_Vulkan completely in the near future.

PR #202, opened 2024-04-05
Shader Extension Proposal

thatcosmonaut:
Tried to create something that would be as minimally invasive to the rest of the API as possible. When using the graphics shader extension, the rest of the API is still used normally, only the Effect calls are omitted.

*Technically* this doesn't break ABI because isRenderTarget on CreateTexture_ is now interpreted as a bitflag, but the value was always 1 or 0 before.


include/FNA3D.h
```diff
@@ -166,7 +166,8 @@ typedef enum FNA3D_CubeMapFace
 typedef enum FNA3D_BufferUsage
 {
 	FNA3D_BUFFERUSAGE_NONE,
-	FNA3D_BUFFERUSAGE_WRITEONLY
+	FNA3D_BUFFERUSAGE_WRITEONLY,
+	FNA3D_BUFFERUSAGE_COMPUTE_EXT /* SDL shader extension only */
```
kg, line 170: Does this mean 'usable by compute AND write only'? Or is it a bit

thatcosmonaut: This value is actually completely ignored on every backend, so setting COMPUTE_EXT on the SDL backend is the only place this would actually matter.

include/FNA3D.h
```diff
@@ -1538,6 +1549,112 @@ FNA3DAPI void FNA3D_SetStringMarker(FNA3D_Device *device, const char *text);
...
+
+/* Binds textures for use with the bound compute pipeline.
+ * You may only call this function after FNA3D_BindComputeShaderEXT has been called.
+ *
+ * pBindings: An array of structs containing texture and cycle option pairs.
+ *            Must be at least as long as the texture bindings specified in the compute shader.
+ */
+FNA3DAPI void FNA3D_BindComputeTexturesEXT(
+	FNA3D_Device *device,
+	SDL_GpuComputeTextureBinding *pBindings
```
kg, line 1629: Is there a blessed way to feed FNA3D textures in to this? Or would you use a SysTexture?
From looking at this it seems like in this proposal, fragment/vertex shaders ('graphics' shaders) can only consume FNA3D textures, and compute shaders can only consume raw SDL_Gpu textures.

thatcosmonaut: FNA3D texture handles _are_ just sdl_gpu texture handles, but I agree that this is awkward... I'll just wrap the SDL struct and handle it in the implementation

thatcosmonaut: Actually I was wrong and FNA3D_Driver_SDL was wrapping the texture handles. So nice catch!

thatcosmonaut:
Now that the GPU proposal is generally approved, I think we can move forward with this.

Upon reflection I think that we shouldn't add a compute extension as part of this. It's a very advanced use case and people who want it will likely benefit more from rewriting their renderer on top of SDL_Gpu. 

What is definitely useful is allowing FNA clients to use modern shader compilers. Some time in the coming months I'll restructure this to only focus on the graphics changes required to let clients pass in shader bytecode and set uniform data outside of the Effect system.

flibitijibibo:
Doing one more check on this - @thatcosmonaut, do you still want to build this or are we firmly moving developers toward SDL3?

kg:
My personal suggestion would be to not do this, and just do the barest minimum amount of SysRenderer and SysTexture integration, since that guides people towards eventually just using raw SDL_GPU for everything.

flibitijibibo:
Making the final call: If you want new shader features, use SDL3 directly.

PR #208, opened 2024-08-27
Draw param order + non-float vertex input fixup

thatcosmonaut:


src/FNA3D_Driver_SDL.c
```diff
@@ -3953,7 +3966,7 @@ static FNA3D_Device* SDLGPU_CreateDevice(
 		MOJOSHADER_sdlGetShaderFormats(),
 		debugMode,
 		0,
-		NULL
+		"Vulkan"
```
flibitijibibo, line 3969:
```suggestion
		NULL
```

PR #211, opened 2024-10-05
Fix verifysamplers crash in SetRenderTargets when no shader is bound

kg:


src/FNA3D_Driver_SDL.c
```diff
@@ -1780,18 +1787,25 @@ static void SDLGPU_VerifySampler(
...
+			{
+				renderer->fragmentTextureSamplerBindings[index].texture = renderer->dummyTexture3D;
+			}
+			else
+			{
+				renderer->fragmentTextureSamplerBindings[index].texture = renderer->dummyTextureCube;
+			}
 		}
-		else
+		else
```
flibitijibibo, line 1806:
```suggestion
		else
```

flibitijibibo (approved):
Just some whitespace

PR #215, opened 2024-10-24
Remove Vulkan renderer

TheSpydog:
Now that we have SDL_GPU, we can remove our old Vulkan renderer. And since it was the only backend that used FNA3D_Memory and FNA3D_CommandBuffer, we can delete those along with it!

Still need to test VisualC-GDK and Xcode, but the other build systems are happy with this change.


flibitijibibo:
Marked as a draft until #219 is done - this will be removed in 25.01 at the latest but if threading support gets done early we could probably do this a month early...

visualc-gdk/FNA3D.vcxproj
```diff
@@ -56,11 +56,6 @@
     <ClInclude Include="..\src\FNA3D_Driver.h" />
     <ClInclude Include="..\src\FNA3D_Tracing.h" />
   </ItemGroup>
-  <ItemGroup>
```
flibitijibibo, line 59: Should these references be removed as part of the move to SDL3?

TheSpydog: Yeah, we need some way of linking with both SDL2/3 (for now) so I'm not sure what this should look like. Though I don't remember making this specific change, the project reference might have been removed automatically by my IDE.

flibitijibibo (approved):
We'll probably want to readd the one GDK thing but other than that we're good to go on this.

thatcosmonaut:
Good night, sweet prince, and flights of angels sing thee to thy rest. 
🪦👼

flibitijibibo:
lol rip

PR #222, opened 2024-12-05
Better diagnostics when an SDL_GPU backend fails to initialize

kg:
Right now if an SDL_GPU backend fails to initialize for any reason you don't get any diagnostic feedback. This should fix that for all the scenarios I can think of.

### DIFF
--- a/src/FNA3D_Driver_SDL.c
+++ b/src/FNA3D_Driver_SDL.c
@@ -4026,7 +4026,13 @@ static void SDLGPU_DestroyDevice(FNA3D_Device *device)
 
 static uint8_t SDLGPU_PrepareWindowAttributes(uint32_t *flags)
 {
-	return SDL_GPUSupportsShaderFormats(MOJOSHADER_sdlGetShaderFormats(), NULL);
+	uint8_t result = SDL_GPUSupportsShaderFormats(MOJOSHADER_sdlGetShaderFormats(), NULL);
+	const char* error = SDL_GetError();
+	if (error)
+	{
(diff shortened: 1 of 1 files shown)


src/FNA3D_Driver_SDL.c
```diff
@@ -4026,7 +4026,13 @@ static void SDLGPU_DestroyDevice(FNA3D_Device *device)
 
 static uint8_t SDLGPU_PrepareWindowAttributes(uint32_t *flags)
 {
-	return SDL_GPUSupportsShaderFormats(MOJOSHADER_sdlGetShaderFormats(), NULL);
+	uint8_t result = SDL_GPUSupportsShaderFormats(MOJOSHADER_sdlGetShaderFormats(), NULL);
+	const char* error = SDL_GetError();
+	if (error)
+	{
+		FNA3D_LogWarn("SDL_GPUSupportsShaderFormats failed: %s", error);
+	}
```
flibitijibibo, line 4034: The error buffer may inherit from a previous call, so we can do the same as the other call here:

```suggestion
	uint8_t result = SDL_GPUSupportsShaderFormats(MOJOSHADER_sdlGetShaderFormats(), NULL);
	if (!result)
	{
		FNA3D_LogWarn("SDL_GPUSupportsShaderFormats failed: %s", SDL_GetError());
	}
```

flibitijibibo (changes requested):
Super tiny fix, otherwise lgtm

flibitijibibo (approved):

PR #227, opened 2024-12-20
SDLGPU: Remove manual fence synchronization

### LINKED ISSUE #225: SDLGPU: Utilize the new swapchain wait functionality
This should let us get rid of the janky fence tracking in SwapBuffers.

thatcosmonaut:
The new `SDL_WaitAndAcquireGPUSwapchainTexture` allows us to avoid manual fence synchronization, if the GPU is falling behind this function will force the CPU to let it catch up.

Fixes #225


thatcosmonaut:
As far as I know an RC for 3.2 is imminent.

PR #228, opened 2024-12-24
Add R8_UNORM and R16_UNORM extended surface formats

kg:
Tested on OpenGL, D3D11 and SDL_GPU.


kg:
You may ask "Why do we need R8_UNORM when Alpha8 exists?" And the answer is that Alpha8 behaves differently on each backend right now, but R8_UNORM as Red is consistent.

flibitijibibo:
If the behavior is different that's likely a bug, unless the APIs define them differently - I would definitely prefer bugs first and extensions dead last, rather than the other way around.

kg:
I looked into fixing Alpha8 before and couldn't figure it out. I'd be happy to pair with someone who wants to look into it I guess? It's also a problem in that Alpha8 RTs don't work in GL, while R8/R16 ones do.

flibitijibibo:
Blocked by https://github.com/MoonsideGames/dav1dfile/issues/4

flibitijibibo:
We're close to the end of FNA's Extension model, but we don't want to block extensions in progress, so I'm doing an approval system... if you get a lot of e-mails, apologies for that! Anyway, for this one:

This one is pretty straightforward but formats tend to have a problem where we don't check the swizzling closely enough and it leads to pretty nasty compatibility breakages, so we _really_ need a working _and publicly available_ example to make sure that we're doing this right and that someone who wants to use this can immediately see how the format is supposed to work.

PR #229, opened 2025-01-01
Implement SysTexture for SDL_GPU backend

kg:
I've tested this locally by using it to create R16_UNORM textures for my dav1dfile test and it worked (with a paired change to FNA to allow me to create a Texture2D instance around a SysTexture handle.)

### DIFF
--- a/include/FNA3D_SysRenderer.h
+++ b/include/FNA3D_SysRenderer.h
@@ -86,6 +86,20 @@ typedef struct FNA3D_SysRendererEXT
 			uint32_t queueFamilyIndex;
 		} vulkan;
 #endif /* FNA3D_DRIVER_VULKAN */
+#if FNA3D_DRIVER_SDL
+		struct
+		{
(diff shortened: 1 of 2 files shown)


kg:
CancelGPUCommandBuffer being commented out is because I can't build locally with those calls in there, the link fails. I'm not sure if it's because something changed on the SDL side or what.

kg:
> CancelGPUCommandBuffer being commented out is because I can't build locally with those calls in there, the link fails. I'm not sure if it's because something changed on the SDL side or what.

This is caused by the way cmake (cmake-gui?) and windows and VS interact, if you don't use the .sln file from your Binaries directory to build you end up with stale lib files. So it's not on FNA3D's end.

flibitijibibo:
SDL's been updated so this should now be clear to rebase.

kg:
Verified again by hand with my 10-bit AV1 player test

PR #240, opened 2025-03-12
Fix sRGB backbuffers on the SDL_GPU and D3D11 backends

kg:
sRGB textures and render targets work but when you request a sRGB backbuffer, you don't get one right now.

On SDL_GPU we aren't flowing the sRGB flag through.

On D3D11, we changed to flip model which doesn't allow sRGB backbuffers, you have to set the colorspace on the swapchain instead.

This PR also fixes a swapchain leak triggered by enabling HDR in D3D11

### DIFF
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -2717,6 +2717,8 @@ static void D3D11_INTERNAL_CreateSwapChain(
 				colorSpace = DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709;
 			}
 			IDXGISwapChain3_SetColorSpace1(swapchain3, colorSpace);
+
+			IDXGISwapChain_Release(swapchain);
 		}
 	}
 
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -2602,7 +2602,9 @@ static void D3D11_INTERNAL_CreateSwapChain(
 		swapchainDesc.BufferDesc.Format = XNAToD3D_TextureFormat[FNA3D_SURFACEFORMAT_COLOR];
 	}
 	else
+	{
 		swapchainDesc.BufferDesc.Format = XNAToD3D_TextureFormat[backBufferFormat];
+	}
 	swapchainDesc.BufferDesc.ScanlineOrdering = DXGI_MODE_SCANLINE_ORDER_UNSPECIFIED;
 	swapchainDesc.BufferDesc.Scaling = DXGI_MODE_SCALING_UNSPECIFIED;
 
--- a/src/FNA3D_Driver_D3D11.c
+++ b/src/FNA3D_Driver_D3D11.c
@@ -2598,6 +2598,7 @@ static void D3D11_INTERNAL_CreateSwapChain(
 	swapchainDesc.BufferDesc.RefreshRate.Denominator = 0;
 	if (backBufferFormat == FNA3D_SURFACEFORMAT_COLORSRGB_EXT)
 	{
+		/* The swapchain RTV uses BGRA8_UNORM but with an SDR Linear colorspace */
 		sRGB = 1;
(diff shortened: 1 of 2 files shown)


src/FNA3D_Driver_D3D11.c
```diff
@@ -2699,6 +2706,10 @@ static void D3D11_INTERNAL_CreateSwapChain(
 			{
 				colorSpace = DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709;
 			}
+			else if (sRGB)
+			{
+				colorSpace = DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709;
```
flibitijibibo, line 2711: Doesn't the `else` cover this?

src/FNA3D_Driver_D3D11.c
```diff
@@ -2595,7 +2596,13 @@ static void D3D11_INTERNAL_CreateSwapChain(
...
 	swapchainDesc.BufferDesc.RefreshRate.Numerator = 0;
 	swapchainDesc.BufferDesc.RefreshRate.Denominator = 0;
-	swapchainDesc.BufferDesc.Format = XNAToD3D_TextureFormat[backBufferFormat];
+	if (backBufferFormat == FNA3D_SURFACEFORMAT_COLORSRGB_EXT)
+	{
+		sRGB = 1;
+		swapchainDesc.BufferDesc.Format = XNAToD3D_TextureFormat[FNA3D_SURFACEFORMAT_COLOR];
+	}
+	else
+		swapchainDesc.BufferDesc.Format = XNAToD3D_TextureFormat[backBufferFormat];
```
flibitijibibo, line 2605:
```suggestion
	{
		swapchainDesc.BufferDesc.Format = XNAToD3D_TextureFormat[backBufferFormat];
	}
```

flibitijibibo (commented):
One tiny thing and one slightly less tiny but still pretty small thing - otherwise lgtm

src/FNA3D_Driver_D3D11.c
kg: I'm not sure this or the other else are both the correct color space

kg: I did some more testing and it seems like they're right, but it feels like either the sRGB or non-sRGB arm should be G10 so I'm not sure what's going on.

src/FNA3D_Driver_D3D11.c
```diff
@@ -2595,7 +2596,15 @@ static void D3D11_INTERNAL_CreateSwapChain(
 	swapchainDesc.BufferDesc.Height = 0;
 	swapchainDesc.BufferDesc.RefreshRate.Numerator = 0;
 	swapchainDesc.BufferDesc.RefreshRate.Denominator = 0;
-	swapchainDesc.BufferDesc.Format = XNAToD3D_TextureFormat[backBufferFormat];
+	if (backBufferFormat == FNA3D_SURFACEFORMAT_COLORSRGB_EXT)
+	{
+		sRGB = 1;
```
flibitijibibo, line 2601:
```suggestion
		/* The swapchain RTV uses BGRA8_UNORM but with an SDR Linear colorspace */
		sRGB = 1;
```

src/FNA3D_Driver_D3D11.c
```diff
@@ -2699,11 +2708,17 @@ static void D3D11_INTERNAL_CreateSwapChain(
 			{
 				colorSpace = DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709;
 			}
+			else if (sRGB)
+			{
+				colorSpace = DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709;
+			}
```
flibitijibibo, line 2714:
```suggestion
```

flibitijibibo (approved):
Compared these changes to SDL_GPU's swapchain composition logic and it's close enough, making a couple changes then merging...

PR #244, opened 2025-03-31
SDL GPU: Intersect scissor with viewport

kiddkaffeine:


src/FNA3D_Driver_SDL.c
```diff
@@ -301,6 +301,24 @@ static inline bool XNAToSDL_PresentMode(
...
 
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
+static inline SDL_Rect computeIntersection(int x1, int x2, int y1, int y2, int w1, int w2, int h1, int h2)
+{
+	SDL_Rect newRect;
+	newRect.x = MAX(x1, x2);
+	newRect.y = MAX(y1, y2);
+	newRect.w = MIN(x1 + w1, x2 + w2) - newRect.x;
+	newRect.h = MIN(y1 + h1, y2 + h2) - newRect.y;
```
flibitijibibo, line 312: SDL_min and SDL_max should cover these macros; we may want to rip off SDL_rect to get a full picture (and catch the weird scenario where there is no intersection):

https://github.com/libsdl-org/SDL/blob/main/src/video/SDL_rect_impl.h#L86

flibitijibibo (changes requested):
The idea makes sense, so we just have to expand on this a bit - let's update the scissor rect whenever the viewport changes too; for example we may have a scissor bigger than the viewport, but when the viewport gets larger we need to tell SDL_GPU that the scissor rect intersection is also larger now.

src/FNA3D_Driver_SDL.c
kiddkaffeine: Changed to SDL_min/max macros, and added a warning when rects do not overlap.

Regarding updating scissor afterwards, it seems like SetRenderTargets actually resets the scissor afterwards:
https://github.com/FNA-XNA/FNA/blob/master/src/Graphics/GraphicsDevice.cs#L1029

If we wanted to sanity check the scissor rect beforehand, wouldn't it be better to do that in the C# side too?

flibitijibibo: Yeah I think you're right - the only place where we update the viewport _without_ updating the scissor is inside FNA3D_SetViewport itself, so I think it'd be a whole separate FNA3D issue if we needed to update the scissor rect while in that function.

flibitijibibo (approved):

PR #246, opened 2025-04-28
Use Debug FNA3D for PowerPC

Fancy2209:
For some reason I haven't figured out, MojoShader can only realize when a Effect file is valid on Big Endian if FNA3D is built as Debug.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,21 +90,15 @@ jobs:
       run: cmake -B debug -G Ninja . -DCMAKE_BUILD_TYPE=Debug -DSDL3_DIR=${GITHUB_WORKSPACE}/SDL/debug
 
     - name: Build (Debug)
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Should we comment the release build temporarily? We can probably strip -S the debug build for uploading as well.

(On the topic, I wonder which -O level it breaks down on?)

Fancy2209:
> Should we comment the release build temporarily? We can probably strip -S the debug build for uploading as well.
> 
> (On the topic, I wonder which -O level it breaks down on?)

Oh yes, commenting it out is a better approach

Also strip was borked iirc, I believe it's a bug in binutils as I haven't been able to strip binaries using cross binutils for PowerPC or PowerPC64, only on actual hardware

Fancy2209:
Ok that's fixed, I should've realized I could've just used powerpc-unknown-linux-gnu-strip

PR #247, opened 2025-05-10
Fix "Format is not compatible with alpha-to-coverage"

kg:
Speculative fix. Haven't tested this locally yet, it was easier to just edit it on github

### DIFF
--- a/src/FNA3D_Driver_SDL.c
+++ b/src/FNA3D_Driver_SDL.c
@@ -1557,6 +1557,7 @@ static SDL_GPUGraphicsPipeline* SDLGPU_INTERNAL_FetchGraphicsPipeline(
 	/* Multisample */
 
 	createInfo.multisample_state.sample_count = renderer->nextRenderPassMultisampleCount;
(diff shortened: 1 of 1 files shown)


kg:
image
image
Seems like my fix is wrong, or doesn't build for some reason.

flibitijibibo:
Let's SDL_zero this structure instead, should work with 3.2 and 3.4+.

kg:
> Let's SDL_zero this structure instead, should work with 3.2 and 3.4+.

Is `struct = {0}` valid in this codebase or should it be SDL_zero specifically

flibitijibibo:
We should do SDL_zero for now, I still need to see if SDL3's language bump includes support for {0}.

flibitijibibo:
Zero'd this specific struct:

https://github.com/FNA-XNA/FNA3D/commit/ce65b2c7d07043558ab1c429e8479ae49b3a0b3f

We should probably do this for all structs with padding but this at least covers the one change that's been made so far.

PR #249, opened 2025-08-01
SDL: Do not stall on uploads

thatcosmonaut:
A partial None buffer update means:
- Overwrite a subregion of the buffer
- Data for preceding draws is unmodified
- Data for subsequent draws uses the new data, and the data outside of the modified subregion is unmodified

This is exactly what a memory barrier accomplishes, so there is no need to stall. 

I think we were also overly zealous with stalling when lots of data is being uploaded. We can just flush the upload command buffer and keep going without stalling, the transfers will begin executing which will free up the internal transfer buffers to be used. Memory usage might go up slightly in the scenario where a lot of small uploads are being performed but it's much better than a hard stall.

Celeste trace looked good to me but I'm marking this as a draft until further testing.

### DIFF
--- a/src/FNA3D_Driver_SDL.c
+++ b/src/FNA3D_Driver_SDL.c
@@ -2988,8 +2988,8 @@ static void SDLGPU_INTERNAL_SetTextureData(
 		}
 		else
 		{
-			/* We already cycled too much, time to stall! */
-			SDLGPU_INTERNAL_FlushUploadCommandsAndStall(renderer);
+			/* We cycled transfers a lot, send the upload commands to reduce further transfer memory usage */
+			SDLGPU_INTERNAL_FlushUploadCommands(renderer);
 			cycle = true;
 			transferOffset = 0;
 		}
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Will run the big traces this weekend but I think the only one that hits this is Murder Miners.

We can also apply this to SetIndexBufferData.

thatcosmonaut:
> We can also apply this to SetIndexBufferData.

Yeah these should just have the same implementation internally.

flibitijibibo (approved):
Did a more thorough review and this should be okay. We may want to write this down inside the driver somewhere, but for now:

The `cycle` parameter refers _only_ to the GPU buffer and _not_ the CPU transfer buffer - the transfer buffer does its own cycling internally and is done regardless of the set mode (including NoOverwrite). The upload to transfer buffer is _always_ a memcpy to fresh memory, so there is no chance of overwrites happening to CPU-visible memory regardless of set mode. At the same time, the CPU->GPU copy _always_ has a memory barrier attached, so NoOverwrite is effectively the same as None, for better or worse (None _finally_ performs well, but NoOverwrite could hypothetically perform even better if we found a way to remove the barrier). Note that this is pretty much the implicit behavior of the OpenGL backend, but _not_ the D3D11 backend; D3D11 still breaks when using None incorrectly!

PR #263, opened 2026-04-26
Add wasm

ValorZard:
removed spriv from emscripten thing to fix build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -385,3 +385,85 @@ jobs:
           sudo pkg install -y cmake ninja sdl3
           cmake -B build -G Ninja .
           cmake --build build --verbose -- -j`sysctl -n hw.ncpu`
+  wasm:
+    name: WebAssembly (${{ matrix.name }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Pthreads
+            artifact_suffix: mt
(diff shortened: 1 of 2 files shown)


TheSpydog:
@ValorZard Can you elaborate on why you had to enable SPIRV support for the WASM build? That seems counterintuitive given that that mojoshader backend is useless on the web.

ValorZard:
@TheSpydog uh it was mostly that if I didn’t add mojo shader to the web build it complained for some reason, and enabling it didn’t seem to cause any problems

ValorZard:
Yeah, without mojoshader, I get this error when running in browser
```
TypeError: can't access property "getParameter", GLctx is undefined dotnet.native.bzw05bjk56.js:8:129877
```

TheSpydog:
I understand that mojoshader is required, but why did you specifically edit the CMakeLists to enable its SPIRV emitter?

ValorZard:
@TheSpydog huh what? I removed it
image

TheSpydog:
MojoShader enables all of its shader emitters by default. In other words, `SUPPORT_PROFILE_SPIRV` defaults to 1. To remove SPIRV support, you have to explicitly opt out via `-DSUPPORT_PROFILE_SPIRV=0`, which is what the CMake file was doing before this change. So what I'm asking is, why did you make this change, which had the result of enabling the SPIRV emitter in mojoshader for web?

PR #264, opened 2026-06-02
SDL: Remove unused line that makes compilation fail

mochaaP:
Fixes: c8fb220d8239baae0e7f7027a4f5aa9d2b0f67e4
Signed-off-by: Zephyr Lykos \<self@mochaa.ws>


flibitijibibo:
Which compiler fails? In what way is this unused? Without any context this just looks like we're leaking something.

mochaaP:
If you do a quick grep, the identifier is nowhere to be found, which probably means a faulty copy paste.
Also SDL_stack_alloc is `alloca`, so SDL_stack_free would be a no-op

PR #268, opened 2026-07-22
you can crash SDL if you allow user resizing and you make the window too small

bartwe:
EllaTAS [C#],  — 14:37
you can crash SDL if you allow user resizing and you make the window too small
Image
 [C#], 
TheSpydog — 15:20
exactly how small is “too small”? this probably isn’t an issue in practice since resizable games should enforce a minimum window size, but still, maybe we’re missing a max() somewhere in the backbuffer resizing logic
bartwe [ROBO],  — 15:21
did a dimension go to 0 ?
bartwe [ROBO],  — 15:50
related to https://github.com/libsdl-org/SDL/issues/10982 perhaps ?
GitHub
GPU: Don't rely on SDL_GetWindowSizeInPixels() for swapchain size  ...
#10958 removed the width and height out parameters from SDL_AcquireGPUSwapchainTexture(). IIRC the arguments for removal were: Out parameters are ugly. The size should always match SDL_GetWindowSiz...
#10958 removed the width and height out parameters from SDL_AcquireGPUSwapchainTexture(). IIRC the arguments for removal were: Out parameters are ugly. The size should always match SDL_GetWindowSiz...
 [ROBO], 
EllaTAS [C#],  — 15:53
yes
EllaTAS [C#],  — 15:54
well, can you set a minimum size in Microsoft.Xna.Framework.GameWindow? I didn't see it
TheSpydog — 16:03
You should be able to clamp it in the ClientSizeChanged callback like so: https://stackoverflow.com/a/45403843
I wonder how XNA handled this case. A backbuffer with 0 width/height is obviously nonsensical
flibitijibibo — 16:07
FNA3D could probably handle it, but i could have sworn we already handled 0 sizes in SDL_gpu_vulkan.c 
 [C#], 
Horse — 16:31
I also got this exact crash myself when testing FNA one week ago by shrinking a resizable window vertically to zero. I only assumed I had to limit the size by myself to avoid the crash 😭
flibitijibibo — 16:40
might be in and out today, but something to try is calling SDL3.SDL.SDL_SetWindowMinimumSize(Game.Window.Handle, 1, 1); which is kind of a sloppy workaround from FNA's point of view but it can be useful for apps that may have a minimum viewport size

### DIFF
--- a/src/FNA3D_Driver_SDL.c
+++ b/src/FNA3D_Driver_SDL.c
@@ -2660,12 +2660,18 @@ static void SDLGPU_INTERNAL_CreateFauxBackbuffer(
 ) {
 	SDL_GPUSampleCount sampleCount = XNAToSDL_SampleCount(presentationParameters->multiSampleCount);
 
+	/* A window can report a size of 0 (e.g. zero-height on Win32), but
+	 * texture dimensions must be >= 1
+	 */
+	int32_t backBufferWidth = SDL_max(1, presentationParameters->backBufferWidth);
+	int32_t backBufferHeight = SDL_max(1, presentationParameters->backBufferHeight);
+
 	if (sampleCount > SDL_GPU_SAMPLECOUNT_1)
 	{
 		renderer->fauxBackbufferColorRenderbuffer = SDLGPU_INTERNAL_CreateTextureWithHandle(
 			renderer,
-			presentationParameters->backBufferWidth,
-			presentationParameters->backBufferHeight,
+			backBufferWidth,
+			backBufferHeight,
 			1,
 			XNAToSDL_SurfaceFormat[presentationParameters->backBufferFormat],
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Makes sense but will wait for the others to chime in since I'm like 99% sure we handled this in SDL_GPU, specifically because Vulkan on Windows tends to spam 0-sized swapchain requests...

TheSpydog:
This looks correct to me. SDL_GPU will handle the scenario where the driver swapchain is 0-sized, but it won't bail you out if you create your own 0-sized texture, which is what's happening with the faux-backbuffer.

From my testing, it looks like the D3D11 backend can also crash in this same way, so we should apply a similar fix there. OpenGL is somehow totally happy with 0-sized renderbuffers though, so at least that's not a concern.

PR #269, opened 2026-08-22
replay: Add support for Emscripten

TheSpydog:
- [x] I confirm that I am the author of this code and release it to the FNA project under the Zlib license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").

Added some CMakery to get Emscripten builds working and split out the per-frame main loop into its own function. The actual logic is basically the same, but all persistent state for the trace (objects, IO, etc) now lives in a static ReplayContext struct.

The indirect return value handling gets a bit wonky, but I didn't immediately see a better way to handle it given that Emscripten main loops can't return error codes (since they never return). Let me know if you can think of a cleaner solution here.

One other small issue is that emscripten requires `BUILD_SHARED_LIBS=OFF`. It'd be nice to not have to specify it when building for web, but I'm not sure if there's a way to change the default on a per-platform basis.

If you want to try this, you can [clone+install+activate the Emscripten SDK](https://emscripten.org/docs/getting_started/downloads.html#sdk-download-and-install) and run:
```
cd SDL
mkdir embuild
cd embuild
emcmake cmake ..
ninja

cd ../../FNA3D
mkdir embuild
cd embuild
emcmake cmake .. -DTRACING_SUPPORT=ON -DBUILD_SHARED_LIBS=OFF -DSDL3_DIR=path/to/SDL/embuild
cp some/random/FNA3D_Trace.bin .
ninja
python -m http.server
```
Then open your browser to the address it serves.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,7 @@ add_definitions(
 	-DFNA3D_DRIVER_OPENGL
 )
 # SDL_GPU does not have a WASM backend, so don't enable it
-if(BUILD_SDL3 AND NOT BUILD_WASM)
+if(BUILD_SDL3 AND NOT EMSCRIPTEN)
 	add_definitions(-DFNA3D_DRIVER_SDL)
 endif()
 add_definitions(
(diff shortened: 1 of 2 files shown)


flibitijibibo (approved):

flibitijibibo:
The returns seem okay to me, worst case we can revisit it as an excuse to migrate to SDL_main callbacks and remove the SDL2 path (for replay, not FNA3D itself).

